PR #6, opened 2015-06-15
Update hwloc to add cache inclusiveness and cpu stepping in topology_…

gbfree:
…hwloc
Modify toplogy_hwloc to get cache inclusiveness and cpu stepping from hwloc.

likwid was using cpuid calls instead of hwloc in topology_hwloc to get the cache inclusiveness and the cpu stepping. Their detection will be added in the next version of hwloc that will be released toward the end of the year. Until then this development version of hwloc can be used.


TomTheBear:
Hi,
thanks for the contribution, looks good. Is the new version of hwloc stable enough, so that we won't get problems until the release end of the year?

gbfree:
Brice Goglin (hwloc main contributor) said that the development version is stable.

TomTheBear:
OK, sounds good, I pull your changes into the master.

PR #8, opened 2015-07-12
Added option to separate install prefix from image directory

gentryx:
Hi, this PR allows package maintainers to configure a different directory for placing files via "make installl" than the executables/libraries will expect at runtime. Most (Linux distribution) packaging tools expect both to be different.


TomTheBear:
Hi,

thanks for your work. I already started implementing this, too.

I have a small question: Why didn't you change every occurrence of PREFIX? As far as I understand it, the PREFIX, BINPREFIX, ... are only used for building while INSTALLED_PREFIX,... are used during installation. In the likwid.lua file, you changed only one path to INSTALLED_PREFIX while two other lines with PREFIX are kept unchanged.

Greetings,
Thomas

gentryx:
Oops, omitting those was an oversight. Fixed.

TomTheBear:
Thanks for fixing the remaining path variables.

I know that chown'ing of files to root is not desirable but before I did that, I got uncountable emails that it does not work for their machine. The main reason was that they simply forgot to set the user of the access and the frequency daemon to root or set the suid bit before changing the user what deletes the suid bit again. Do you know a solution for this? It was not enough to document it.

gentryx:
Got it. I added an option to config.mk so packagers can disable it, but for users chowning to root will stay on. Would that be acceptable for you?

TomTheBear:
OK, thanks. That's fine for me. I merge the branches now.

gentryx:
Cool, this is highly appreciated. Could you push out a 4.0.1 release? I'd then update the Gentoo ebuild. ATM all official Gentoo packages for likwid are broken and I'd be happy to get this fixed. :-)

TomTheBear:
I planned to do it this week. I have to do some tests before I can release it.

PR #9, opened 2015-07-16
 topology_hwloc : debug cache inclusiveness detection with some core/PU restricted by cgroups

gbfree:
Hwloc only annotate the inclusiveness of the cache detected by his x86 backend. It can't detect cache without binding to his PU. 
If at a level the cache detected as 0 by the linux backend has no bindable pu his inclusiveness will be null.
As likwid assume that all the cache at one level are the same, cousin inclusiveness can be used instead.

May fix https://groups.google.com/forum/#!topic/likwid-users/XD6W1z203wE
(unless additional information are provided and cgroup weren't used)


TomTheBear:
Thanks for the fix, I merge it to the master branch.

We will see whether it fixes the problem reported on the mailing list.

PR #20, opened 2015-11-06
Add tests to .travis.yml

TomTheBear:


TomTheBear:
Virtual machine do not properly implement the MSR registers. So no perfmon testing at travis-ci

PR #25, opened 2015-11-10
Make error message a bit more verbose

Shurakai:
Let's tell the user that he might have to adjust the PATH environment variable...


TomTheBear:
Thanks for your commit but the message needs to be further extended. The message is printed when neither the access daemon folder is in the PATH variable nor at the location given during build process in the config.mk file.

PR #40, opened 2016-06-20
Fix for likwid-powermeter, which uses the wrong RAPL domain when requesting energy

rschoene:
Simple fix, just pass the correct domain, not zero.


TomTheBear:
Thanks for the fix. I just read the power related code after your issue and didn't find this one. Thanks.

PR #51, opened 2016-08-31
Fixed the incorrectly named events in the PORT_USAGE group on sandy/ivybridge

K-Wic:
Change the UOPS_EXECUTED_PORT in the PORT_USAGE group to UOPS_DISPATCHED_PORT for sandybridge and ivybridge. Also limited the number of ports to the available six.


TomTheBear:
Thanks for the correction.

PR #52, opened 2016-08-31
Turning on the debug build now also disables optimizations for GCC.

K-Wic:
Not using -O0 is a real pita when trying to debug things with gdb.

Unfortunately this step also requires uninlining two functions, which are optimized away in the release build, but cause linker problems when left in.


TomTheBear:
Hi, thanks for the PR. I always debugged with -O2 and it was suitable for my needs. I won't merge your PR because I don't want to have general changes in the compiler-specific files. I added it to config_defines.mk where the -g flag is added to CFLAGS when DEBUG is set to true. The inlines are also removed from the two functions in calculator.c

PR #67, opened 2016-11-03
LIKWID support for POWER8 processors.

TomTheBear:
This version of LIKWID supports IBM POWER8 processors. Thanks to the IBM Labs @ Boeblingen for their support.


TomTheBear:
This PR is outdated. It uses a native backend to access the performance registers. On the one hand LIKWID can do everything itself with less overhead but on the other hand you have to load a kernel module which enables the user-level access to the registers.
Therefore, I added another PR which includes POWER8 and POWER9 and uses perf_event as backend. It's not 100% stable but I'm working on it.

TomTheBear:
I close this PR. The support for POWER8 and POWER9 was merged into master from PR  #217 .

PR #68, opened 2016-11-11
Fix event options

K-Wic:
Fixed several EVENT_OPTIONS that did not match their corresponding events.


TomTheBear:
Thanks for fixing the groups and event options.

PR #79, opened 2017-02-09
Implement Skylake frontend events

mgottschlag:
While the frontend events were defined, the MSR_PEBS_FRONTEND register apparently wasn't configured anywhere.


TomTheBear:
Thanks for the PR but after I rechecked the documentation of the FRONTEND_RETIRED events. It seems that PEBS is required to use these events and that's not possible for LIKWID. PEBS requires to set up some kernel-level buffers which cannot be done by LIKWID as it runs completely in user-space.

My opinion: In order to avoid confusion the FRONTEND_RETIRED events should be deleted from the event lists.

mgottschlag:
In "Table 19-3. Non-Architectural Performance Events of the Processor Core Supported by Skylake Microarchitecture", the events are marked "PS", but "PS" is defined as "Also supports PEBS." (where "also" implies that they do not require PEBS). While the detailed description is strongly linked to PEBS, that table isn't.

In fact, the events seem to work on real hardware, without any kernel code.

TomTheBear:
OK, I have never tested them. I just checked the event definitions and there are inconsistencies between Intel SDM and the Perfmon online DB. You PR seems fine and I'll merge it soon, thanks for your contribution.

PR #80, opened 2017-02-14
set correct path to liblikwidpin

jeorsch:
For LIBLIKWIDPIN the wrong INSTALLED_PREFIX instead of the correct INSTALLED_LIBPREFIX is used to determine the path to the shared object. Fixed this in config.mk.


TomTheBear:
Thanks for the PR

PR #82, opened 2017-02-20
Fixed error in Sandy Bridge L3 miss ratio calculation

LangdalP:
I noticed that on my Sandy Bridge rig, the L3 miss ratio did not range between 0-1, but between 0 and some arbitrarily high number. This was because the division failed because it was wrongly specified.

I do not fully understand the syntax used for the calculations, but I managed to fix it by changing the divisor to `PMC1:MATCH0=0x0081:MATCH1=0x1`, which matches how it was specified in the EVENTSET section. Before I changed it, the divisor was `PMC1:MATCH0=0081:MATCH1=0x1`. Notice that `MATCH0` is set to 0081 before, and 0x0081 after.

After making this change, the L3 miss ratio is equal to `OFFCORE_RESPONSE_0_OPTIONS` / `OFFCORE_RESPONSE_1_OPTIONS`, which seems to range between 0-1 in my experiments.

As mentioned, I do not fully understand the file syntax, so someone should inspect the changes to see that they are correct.

What was done, specifically:
'MATCH0=0081' was changed to 'MATCH0=0x0081' in a few places


TomTheBear:
Thanks for the PR. Your changes affect only the group file for desktop SandyBridges (groups/sandybridge/L3CACHE.txt) but the same error exists in the server SandyBridges files (groups/sandybridgeEP/L3CACHE.txt). Can you change this file, too, so that the PR is complete?

LangdalP:
No problem. I have now applied the same fixes to that file.

PR #89, opened 2017-03-15
Use SLURM_* vars when no MPI_LOCAL* vars

aaronknister:
The environment variables MPI_LOCALNRANKS and MPI_LOCALRANKID don't appear to be native to SLURM. So as to not break sites alread using likwid that may have these MPI_LOCAL* vars use the SLURM_* equivalents only if the MPI_LOCAL* vars are unavailable.


TomTheBear:
Looks reasonable, can be merged.

Does the SLURM support work for you? I tested it only on one system, so there might be errors on other systems.

aaronknister:
Yep, it worked for me with a couple tests I ran. Thanks!

TomTheBear:
OK, thanks for the information.

PR #95, opened 2017-04-11
Add missing return statements for the perf_event support.

mcolmant:
Some return statements are missing when using the perf_event support.

To be compliant with the other headers, I added the same statements where they were missing.


TomTheBear:
Thanks for trying and refining the perf_event backend.

Issue #96, opened 2017-04-11
Question about the perf_event support


mcolmant:
Hi,

First of all, I would like to thank you for this awesome library and for your reactivity!
I have one question regarding the perf_event support.
I have a specific need regarding this support.
As you may know, the file descriptor are opened with the `perf_event_open` function while using this interface.
```
int perf_event_open(struct perf_event_attr *attr,
                           pid_t pid, int cpu, int group_fd,
                           unsigned long flags);
```
I already looked into the code and this function is wrapped inside `perf_*_setup` functions.
What I would like to do is to adapt the code for allowing to use of the `pid` and `flags` parameters.
The main modification will be located to the `perfmon_setupCountersThread`, but will lead to a lot of impacts.
Maybe you will have some advices to limit this impact.

I already wrote something, are you interested by such feature?

TomTheBear:
Hi,

In general, I have no problem when perf_event is "limited" to a PID and the flags are set as long as it is configurable. The only concern I have is that it changes the measurement behavior. Normal LIKWID measures anything that runs on the selected cores. Perf_event with PID measures only PID and, if activated, the child processes. 

I would recommend to define new event options for PID and flags when compiled with perf_event support. Consequently, you need a loop in perfmon_setupCounterThread to search for the options and use them for the subsequent perf_event_open call. This limits the changes basically on the perf_event backend.

mcolmant:
Hello,

Thanks for your quick answer!
I understand what you mean. 
I agree, it will be more elegant this way :-)
I'm working on it, and will come back with a PR (if you are interested for such feature in likwid).

TomTheBear:
Hi,
you're welcome. Thanks for your attempt to enrich LIKWID. 
I'm very interested, it is always helpful to support more measurement backends. If you are experienced in the use of perf_event, it would be helpful to check the uncore support. I'm not really sure whether all the code is working for the uncore units, I only checked a few units and events.

mcolmant:
I don't have the time to test everything right now, but if I find something, I will let you know.
Right now, it seems to work well.

mcolmant:
Hello, 

One last question. I would like to use likwid (with this new support) to monitor several processes at runtime.
It seems that likwid does not support this type of monitoring (`perfmon_perfevent.h` uses a struct that allows to only have one counter per core).
[...]

PR #98, opened 2017-04-12
Add the support of the pid and flags parameters (perf_event).

mcolmant:
Add the possibility to use the PID and flags parameters for the perf_event_open call when using the perf_event interface (#96).

### DIFF
--- a/src/includes/perfmon_perfevent.h
+++ b/src/includes/perfmon_perfevent.h
@@ -310,10 +310,29 @@ int perfmon_setupCountersThread_perfevent(
         }
         if (ret == 0)
         {
-            cpu_event_fds[cpu_id][index] = perf_event_open(&attr, 0, cpu_id, -1, 0);
+            pid_t pid = 0;
+            unsigned long flags = 0;
+
+            for (int j = 0; j < event->numberOfOptions; j++)
+            {
+                switch (event->options[j].type)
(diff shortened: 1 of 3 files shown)


mcolmant:
Oh, thanks, it has been fixed.

TomTheBear:
Looks fine and my first tests don't throw any errors.
What came into my mind: If you want to use the PID feature for the wrapped executable, we have to reconstruct the flow of likwid-perfctr. Currently, the eventset is added to LIKWID and the events are set up before the executable is started and you get the PID. I would require something like, start the executable and pause it directly. Afterwards get the PID, add the eventset with PID option, set up the events, start the counters and unpause the executable.
Or is your use-case always that the executable is already running and you want to monitor it from the outside?

mcolmant:
For what I want to do right now, I'm monitoring the PID from outside. 
But I have already used a similar approach (start, pause, counters, restart) to use the inherit option (child processes).
I think it could be good to add this support also in the wrapped executable, maybe it can be useful for other users.

TomTheBear:
I implemented the start/pause/counters/unpause scheme in likwid-perfctr. Moreover, a command line option "--execpid" that adds perf_pid=<PID_of_the_started_application> to the last event in the event string. So, I think this PR can be merged.

PR #122, opened 2017-10-12
AMD Zen: Don't invert thread/slice selectors

lluchs:
Is there any reason why these are inverted? If I'm trying to filter events by thread, I usually think of "I'd like to get events from *these* threads", not "I'd like to **not** get events from these threads".


TomTheBear:
I'm not sure why I inverted the mask but I created the Zen support without a system. I just tried it on our Zen machine and you are right, the masks shouldn't be inverted. Thanks for finding and fixing it

PR #124, opened 2017-10-30
Fixed double `free`

lunixoid:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. Error, found using PVS-Studio:
likwid/src/calculator.c     771     err     V586 The 'free' function is called twice for deallocation of the same memory space. Inspect the first argument. Check lines: 760, 771.


TomTheBear:
Many thanks for the PR.

PR #148, opened 2018-02-09
added ansible role

yogeshVU:
added ansible role which currently dowloads stable release from the ftp server and installs the stable release on the target server.


TomTheBear:
Is it required by ansible to have its folder directly at first level? Because ansible is not the only tool for which people asked to have packaging stuff (spec-File for RPM, Debian rules, ...) and it would be nice to have it grouped together in some subfolder.

yogeshVU:
@TomTheBear ansible can be used for targeting various distributions(ubuntu/centos..etc). For now I have it tested only on ubuntu. 
Directory structure placement does not matter for the ansible. It can also be moved to a totally different repository if required.

TomTheBear:
Thanks for the clarification. What I had in mind is to put the ansible stuff in a subfolder like 'packaging' or similar, so that we can group all packaging stuff there.

yogeshVU:
That's a good approach.

yogeshVU:
Any changes required to make this merged?

TomTheBear:
I'm currently preparing a new 4.3 release, so only little time to take care of the master branch.

You could move the 'ansible.likwid' folder to a subfolder 'packaging' and rename your folder in 'ansible'. I still havn't had time to look more deeply into ansible but since it builds from source, is it possible to make configuration steps (like changing the access mode or similar)? Commonly, these packages contain information like a description, license, etc. Is this information already included in your commit or should it be added after merging?

Issue #153, opened 2018-04-29
Replace perl she-bangline


fendor:
On some Unix variations, perl is not installed in `/usr/bin/perl`, e.g. in NixOS.
This forced me, to manually replace all occurrences of `#!/usr/bin/perl` to `#!/usr/bin/env perl`.
Then I was able to successfully compile the likwid tools.

Would it be possible to replace `#!/usr/bin/perl` in upstream with `#!/usr/bin/env perl`?
It should have the same functionality while simplifying the installation process for some Operating Systems and likwid is then, overall, more portable.

TomTheBear:
Thank you for the PR. The perl scripts havn't been touch for a long time.

PR #154, opened 2018-04-29
Replace perl she-bangline 

fendor:
This is a pull request for the issue #153.


fendor:
The second commit is to add warning flags to the perl files.

TomTheBear:
Thank you for the PR. Looks good to me.

PR #180, opened 2018-08-09
fix likwid-mpirun bug with slurm

ohlmann:
- the problem lies in setting SLURM_NODELIST in likwid-mpirun
- currently, the hostlist is read from a scontrol command using a lua
  pipe with f:read('*a')
- this leads to line break at the end of the environment variable
  SLURM_NODELIST, causing srun to segfault
- this can be fixed by using f:read('*l') because the scontrol output
  should always be only one line -> the line break is ignored

### DIFF
--- a/src/applications/likwid-mpirun.lua
+++ b/src/applications/likwid-mpirun.lua
@@ -572,7 +572,7 @@ local function writeHostfileSlurm(hostlist, filename)
     cmd = string.format("scontrol show hostlist %s", table.concat(l,","))
     f = io.popen(cmd, 'r')
     if f ~= nil then
-        likwid.setenv("SLURM_NODELIST", f:read('*a'))
+        likwid.setenv("SLURM_NODELIST", f:read('*l'))
         f:close()
     else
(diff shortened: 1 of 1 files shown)


ohlmann:
For me, this seems to fix one of the issues reported in https://groups.google.com/forum/#!topic/likwid-users/Qd1vpfhL3qg (srun did not work in the script because of the different environment variables).

ohlmann:
I think so, because the hostlist expression is always one line (e.g. broadwell[001-009,014],skylake01).

TomTheBear:
I checked the SLURM documentation and it's not directly mentioned but you can somehow read it out that it always on one line:
> If no hostlist expression is supplied, the contents of the SLURM_JOB_NODELIST environment variable is used. For example "tux[1-3]" is mapped to "tux1","tux2" and "tux3" (one hostname per line). hostlist takes a list of host names and prints the hostlist expression for them (the inverse of hostnames)

The inverse of "one hostname per line "would be "all hostnames in one line".

PR #183, opened 2018-09-11
Events syntax

termim:
Some typos in events descriptions


TomTheBear:
Thanks for checking and fixing the events.

PR #184, opened 2018-09-12
Events syntax

termim:
More fixes


TomTheBear:
Thank you again. Nice work and I should be more careful when adding new events

PR #188, opened 2018-11-05
Update likwid-mpirun.lua

flnowak:
Function executeOpenMPI now also initializes bindstr when using Open MPI v3.x.x


TomTheBear:
Thank for the PR. I didn't even know that OpenMPI reached version 3.x already. I think this PR can be merged.

The travis-ci message is misleading as it fails setting up the base environment. In this case it cannot find a suitable package for gcc.

flnowak:
Thank you too for checking the PR this fast!

PR #194, opened 2018-12-06
Do not restrict access daemon to single CPU

hahnjo:
I think this comes from a time where likwid would start one daemon
per CPU. Nowadays there is only one daemon (unless you are in a
threaded environment?) and consequently reading hardware registers
becomes really slow if there is a user application also pinned to
that CPU.

### DIFF
--- a/src/access_client.c
+++ b/src/access_client.c
@@ -135,13 +135,6 @@ access_client_startDaemon(int cpu_id)
 
     if (pid == 0)
     {
-        if (cpu_id >= 0)
-        {
-            cpu_set_t cpuset;
-            CPU_ZERO(&cpuset);
-            CPU_SET(cpu_id, &cpuset);
-            sched_setaffinity(0, sizeof(cpu_set_t), &cpuset);
-        }
         ret = execve (exeprog, newargv, newenv);
 
         if (ret < 0)


TomTheBear:
In the MarkerAPI case, where multiple threads require to access the hardware registers simultaneously, multiple AccessDaemons are started. I'm not sure how much overhead it creates when a thread running on core X communicates with a daemon on core Y that accesses MSRs at core X. But the idea was to avoid it and keep the daemon on the same core as the requesting thread.

In cases where one thread/process reads all cores, only one daemon is started. In this case it is pinned to the first CPU in the CPUset. This is indeed not required.

During the MarkerAPI initialization phase only one daemon is started and currently pinned to the first CPU in the eventset (the one the master thread runs on). When registering or starting the first region, LIKWID starts the remaining daemons because it detects a multi-threaded environment. The daemon for the master thread would still have the whole CPUset when not pinned in the first place but this could be fixed by changing the CPUset for the master daemon when detecting a multi-threaded environment.

hahnjo:
If the MarkerAPI is the main use case, would it be an option to disable binding unless started with the MarkerAPI? If not how would you propose to solve the issue? I think this is what's causing wrong number in our monitoring system because likwid can't read the hardware counters fast enough...

TomTheBear:
I committed and improved version of your PR. Common runs start daemon without pinning, MarkerAPI/multithreaded pins the daemon as soon as detected
https://github.com/RRZE-HPC/likwid/commit/01487499c631ce2b1ed59076b3dc6f5770051fac

Issue #199, opened 2019-01-31
provide python functions to post-process/parse terminal output


davydden:
for collecting data and plotting, it would be useful to have some auxiliary python functions (as part of the likwid installation) that one can use to parse the terminal output (i.e. regions, groups, metrics). 

Alternatively, in addition to the terminal output maybe likwid could write such data in some format readable by python, say write `sqlite3` database.

EDIT: there are a few options to help parsing in python ([here](https://penandpants.com/2012/03/09/reading-text-tables-with-python/))

EDIT2: I wrote a parser, which seems to be doing the right thing. If you are interested and tell me where to put this, I can create a PR.

davydden:
LIKWID has `-O` option to output in `CVS`, so that should be the fastest and robust route to get data into python. I simply was not aware of this flag.

PR #200, opened 2019-02-03
python: add terminal output parser of tables

davydden:
fixes https://github.com/RRZE-HPC/likwid/issues/199

@phibel I checked this with `likwid-mpirun` and `MEM_DP` group with markers, but I know for sure in it's current form it won't work for `likwid-perfct` as it has a different output format for regions and groups. Hopefully this script can be improved/bended to cover that case, as well as `likwid-powermeter`. Feel free to create PRs to this branch.

### DIFF
--- a/python/utilities.py
+++ b/python/utilities.py
@@ -1,0 +1,105 @@
+import os
+import fileinput
+
+def parse_likwid_file(filename, last_line = ''):
+    '''Parse terminal output of LIKWID
+
+    Arguments:
+    filename -- the path to the file
+    last_line -- the last line of non-LIKWID output. Starting from the line after this
+        the parser will be active
+
+    Return:
+    parsed data stored as:
+        dictionary (region)
+        dictionary (table name i.e. Event, Event_Sum, Metric, Metric_Sum or alike
+        dictionary of rows.
+
+    That is, one can get the row as result['vmult']['Metric_Sum']['MFLOP/s']
+    '''
+    result = {}
+    fin = open(filename, 'r')
+    debug_output = False
+
+    row_separator = '---------'
+
+    found_start = (last_line == "")
+    region = ''
+    separator_counter = 0
+    table_name = ''
+    table_ind = 0
+
+    for line in fin:
+        line = line.strip()
+
+        # skip empty lines
+        if line == "":
+            continue
+
+        if found_start:
+            #
+            # Main logic to parse
+            #
+
+            # Check if we found one of the regions:
+            if 'Region:' in line:
+                region = line[8:]
+                if debug_output:
+                    print '-- Region: {0}'.format(region)
+                # regions should be unique
+                assert region not in result
+                result[region] = {}
+                separator_counter = 0
+                table_name = ''
+                continue
+
+            if 'Group:' in line:
+                # FIXME: read in groups as well?
+                continue
+
+            # If we are in LIKWID part, we should have some region always around
+            assert region != ''
+
+            # At this point we have only 3 options: we are on one of the separators,
+            # inside the header or inside the core of the table.
+            if row_separator in line:
+                separator_counter = separator_counter + 1
+                # reset the counter if we are at the end of the current table
+                if separator_counter == 3:
+                    separator_counter = 0
+                continue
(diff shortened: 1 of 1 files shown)


davydden:
@TomTheBear I looked at the `CVS` output, although it's easier to parse, it's still not user friendly in that one needs to manually separate data for different tables (specific regions, sum or events, etc). 

Looking at the output, I have a feeling that it is **not** easier to parse because the output completely lacks any separators of actual tables. Namely it is just a single huge _table_ (actually not):

```
Region,blabla,
Group,1,
Event,Counter,e0107:0:0
Runtime (RDTSC) [s],TSC,1.205799
Region calls,CTR,10
INSTR_RETIRED_ANY,FIXC0,5067253000
...
CAS_COUNT_RD,MBOX7C0,0
CAS_COUNT_WR,MBOX7C1,0
Metric,e0107:0:0,
Runtime (RDTSC) [s],1.2058,
...
Memory data volume [GBytes],9.5893,
Operational intensity,1.0240,
Region,BLABLABLA,
Group,1,
Event,Counter,e0107:0:0
...
```

One could of course add some keywords, i.e. we know that `Region` and `Group` are NOT tables.
But then I would need to add all possible keywords for other table headers, i.e. `Event` and `Metric`. 
So I am not yet sure if it's easier to parse CVS or terminal output. Terminal output at least has separators for start/end of the actual table:
```
+--------------------------------------+---------+-------------+
|                 Event                | Counter |  e1109:0:0  |
+--------------------------------------+---------+-------------+
BLABLA
+--------------------------------------+---------+-------------+
```
so it's easier to catch the header of the table and the actual content without extra assumptions. 

Bottom line is, I think python parser is still userful as it allows one to work with parsed results via
```
result['vmult']['Metric_Sum']['MFLOP/s']
```

For my needs, I would probably stick with parsing the beautified terminal output.

TomTheBear:
Hi,

I understand the need of such a parser but I don't see your arguments against the CSV output. It actually contains keywords already (STRUCT and TABLE):
```
STRUCT,Info,3,,,
CPU name:,Intel(R) Core(TM) i7-4770 CPU @ 3.40GHz,,,,
CPU type:,Intel Core Haswell processor,,,,
CPU clock:,3.392181359 GHz,,,,
TABLE,Region copy,Group 1 Raw,L3,5,
Region Info,Core 0,Core 1,,,
RDTSC Runtime [s],0.490816,0.461983,,,
call count,10,10,,,
Event,Counter,Core 0,Core 1,,
INSTR_RETIRED_ANY,FIXC0,602320300,600771400,,
CPU_CLK_UNHALTED_CORE,FIXC1,1503458000,1498534000,,
CPU_CLK_UNHALTED_REF,FIXC2,1503456000,1498554000,,
L2_LINES_IN_ALL,PMC0,49862770,50043380,,
L2_TRANS_L2_WB,PMC1,15348130,15296480,,
TABLE,Region copy,Group 1 Raw STAT,L3,5,
[...]
```

davydden:
@TomTheBear 

> It actually contains keywords already (STRUCT and TABLE):

which version of LIKWID is that? I don't have those keywords in the output from 4.2.1.

TomTheBear:
A more recent one (4.3.3) ;) The markers in CSV were added in May 2015 (https://github.com/RRZE-HPC/likwid/commit/ea60dda3f60a6724601232e571510a6ffc1a1ec2)

davydden:
> A more recent one (4.3.3) ;) The markers in CSV were added in May 2015 ([ea60dda](https://github.com/RRZE-HPC/likwid/commit/ea60dda3f60a6724601232e571510a6ffc1a1ec2))

👍 

ah, ok, let me try with the most recent version and see what native python parsers do with such output...

davydden:
@TomTheBear 4.3.3. is not there on Emmy, whereas 4.3.2 does NOT contain keywords on CVS output. At least not from 
```
likwid-mpirun -np 1 -nperdomain S:10 -g MEM_DP -m -O
```

I will until 4.3.3. is available to see how CVS works and if it's easier to parse. Will keep the PR open for now...

PR #205, opened 2019-03-03
Check data

termim:
Add script for checking event files grammar and event/subevent names consistency.
Fix typos and inconsistencies discovered by the script.
There are still several grammar errors not fixed - I couldn't find corresponding event codes in Intel docs.


TomTheBear:
Wow, I'm impressed. Thanks! It would have take weeks for me to recheck all event files.

What do you mean with

> There are still several grammar errors not fixed - I couldn't find corresponding event codes in Intel docs.

There are some events in the event lists that have been added by me for convenience and completeness. There are two kinds of these events:

- If there exist UMASKs that seem to correspond to each other (e.g. evicts of clean cache lines and evicts of dirty cache lines) I derive a combined event (e.g. all evicted cache lines) by OR'ing the UMASKs. The event names are chosen by me.
- Some events exist in the documentation for a previous microarchitecture but are not listed anymore for the current microarchitecture although it is known that there have been only slightly changes (e.g. SandyBridge/IvyBridge or Haswell/Broadwell). In some cases I try the event of the "old" generation on the "new" generation and if it works (with reasonable counts, sometimes validated with likwid-bench), I add them to the list with a suitable name, either the original name of the "old" architecture or a self-chosen one to fit the naming scheme of the "new" architecture.

The first case can probably be processed by your script but the second one is probably a lot of work. It would be probably easier to have some files with the special cases. I thought about getting the event list from the web at compile time but I would need additional files for the self-defined events. Unfortunately, just Intel provides the events in a machine readable format at an easily downloadable location.

TomTheBear:
Should I merge or do you want to add more stuff to the PR?

termim:
> > What do you mean with
>     There are still several grammar errors not fixed - I couldn't find corresponding event codes in Intel docs.

There are only four cases actually. You'd see them if you run 
`python3 check_events_files.py events`
They all about missing subsystem name. I'd guess that the subsystem is PMC in those cases but
I couldn't find event code in Intel's SDM. I could just miss them though. See for example:

`
perfmon_nehalem_events.txt:206:40:
Expected pipe (|) separated list of subsystem names (PMC, PBOX0|PBOX1 etc.):
EVENT_UOPS_DECODED                 0x3D`

> Should I merge or do you want to add more stuff to the PR?

If you find the changes meaningful then yes.

As a next step I'd do something similar for the group files.

TomTheBear:
That's true, the event `UOPS_DECODED` is not well defined in the event file. Like you, I assume PMC for the event. But I cannot check it anymore because it is not listed in any of my documentation. But is was already present in LIKWID 2.x (also without counter definition), so I asked the former developer where he found the event and he will check the old documentation.

Of course the script and fixed event files are meaningful. I just wanted to wait whether you want to add more commits to this PR. 

Thanks for considering the group files. There is already a script `test/check_group_files.py` but compared to your event file script, the script is more basic.

TomTheBear:
I just ran your script to fix the warnings (except Nehalem UOPS_DECODED) and found this in perfmon_zen_events.txt:
```
EVENT_RETIRED_CLFLUSH                   0x26 PMC
UMASK_RETIRED_CLFLUSH                   0x00

EVENT_RETIRED_CPUID                     0x26 PMC
UMASK_RETIRED_CPUID                     0x00
```
Both use the same event ID, which is wrong. EVENT_RETIRED_CPUID has ID 0x27 but I don't know how to detect this in a reasonable way because there are cases where it should be like this: same event ID but different names (commonly in combination with some default options).  An example for this is EVENT_CYCLE_ACTIVITY for Skylake/Haswell/Broadwell.

PR #212, opened 2019-03-26
Moved initialization of vectors into bench.c:runTest()

### LINKED ISSUE #210: Implement per thread first-touch policy in likwid-bench
`likwid-bench` allocates and initializes data on first thread, not on the threads which will eventually use it. This leads to performance issues if the thread group in a domain spans multiple NUMA domains.

It should be sufficient to move initialization with the threads for better performance and more intuitive behaviour.

cod3monk:
This resolves #210

### DIFF
--- a/bench/src/allocator.c
+++ b/bench/src/allocator.c
@@ -161,48 +161,5 @@ allocator_allocateVector(
             offset,
             LLU_CAST elements);
 
-    switch ( type )
-    {
-        case INT:
-            {
-                int* sptr = (int*) (*ptr);
-                sptr += offset;
-
-                for ( uint64_t i=0; i < size; i++ )
-                {
-                    sptr[i] = 1;
-                }
-                *ptr = (void*) sptr;
-
-            }
-            break;
-
-        case SINGLE:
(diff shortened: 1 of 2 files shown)


TomTheBear:
Generally, the fix should work but I think it breaks the feature to specify the location of streams like `-t copy -w S0:100MB:1-0:S1,1:S1` where threads run in domain S0 and both input and output streams are located at S1. The thread in allocator.c get pinned to domain->processorList[0] => S1 while in bench.c it is myData->processors[threadId] => S0.

TomTheBear:
Moreover, the initialization of the streams is now included in the fallback timer (https://github.com/RRZE-HPC/likwid/blob/master/bench/likwid-bench.c#L505-L508). In most cases it uses the CPU cycles as base for timing but some archs (like ARM) do not provide a simple cycle counter and uses this fallback timer.

cod3monk:
From a usability point of view, I would expect first-touch to be used if no stream locations are specified. Would you agree to that? If so, both implementations could be used and selected depending on the `-w` string.

PR #217, opened 2019-04-03
LIKWID support for POWER8 and POWER9 systems

TomTheBear:
This version of LIKWID supports IBM POWER8 and POWER9 processors. Thanks to the IBM Labs @ Boeblingen and Carl Love from IBM the for their support. Thanks to TU Dresden to provide a test system. This version uses perf_event as backend, so no kernel driver is required like in the other "POWER8 pull request"


frankicia:
When trying to compile it, I get the following error:

```
./src/access_x86_msr.c: In function ‘test_rdpmc’:
./src/access_x86_msr.c:83:5: error: impossible register constraint in ‘asm’
     __asm__ volatile("rdpmc" : "=a" (low), "=d" (high) : "c" (counter));
     ^~~~~~~
make: *** [GCC/access_x86_msr.o] Error 1
```

GCC used is 6.4.0. OS is Red Hat Enterprise Linux Server 7.5

TomTheBear:
You have to switch the compiler setting in config.mk to GCCPOWER. This filters out the x86- and arm-specific files.

frankicia:
Yes, I realized that.
With GCCPOWER there is a small error in file src/includes/perfmon_perfevent.h 
Line 83:
#if defined(__x86_64__) || defined(__i386__) || define(_ARCH_PPC)
should be
#if defined(__x86_64__) || defined(__i386__) || defined(_ARCH_PPC)

TomTheBear:
For POWER9, there are still devices not supported by this PR. I have to see where I get enough documentation to integrate them into LIKWID.

PR #222, opened 2019-05-31
Changing type of variable 'offset'

dkuts:
Changing type from 'int' to 'size_t' to avoid overflowing.
At some cases, offset may be bigger than 2 147 483 647, that cause negative number and access to wrong memory address. 
example:
likwid-bench -t copy_avx -w S0:48GB:48 −i 2048

### DIFF
--- a/bench/src/bench.c
+++ b/bench/src/bench.c
@@ -135,7 +135,7 @@ runTest(void* arg)
 
     /* pin the thread */
     likwid_pinThread(myData->processors[threadId]);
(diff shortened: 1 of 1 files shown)


TomTheBear:
Thanks for the PR. There are more locations where we should do adjustments (printf in line 138 throws warnings with %d format and size_t variable, function getIterSingle() uses an int offset as well). How you want to do it? Do you want to integrate all changes in your PR or should I merge your PR and do the other changes myself?

dkuts:
OK, I've fixed other usage and printf format.

TomTheBear:
Thanks for the update. Ready to merge.

PR #233, opened 2019-07-20
Group syntax

termim:
Please consider the following changes:

- add syntax checker for group files
- fix syntax/typo errors in the group files


TomTheBear:
The checker is quite strict but it's fine like that. For example, LIKWID requires the LONG section to be non-empty but all content is up to the group creator. So, if he/she doesn't want the formulas shown in the documentation section (LONG), that's fine. I always add the formulas, so that's why it is fine for me.
So, from my side it is ready to merge.

termim:
Please let me know if you want to relax this check or make it optional.
Actually the 'Formulas' part is somewhat redundant as it could be automatically produced from 'METRICS' and 'EVENTSET'. I was thinking even to verify formulas as I saw a discrepancy between formula and eventset somewhere.

TomTheBear:
I'm fine with the current handling in your parser. Formula verification would be good, though. I do copy&paste most of the time for the base skeleton and forget to adjust the formulas sometimes.

I already thought about a "group creator" where you add just a short description, eventset, metrics and the long documentation (without formulas) and the remaining stuff is added automatically but you still have to do the main work, so I thought adding the formulas manually doesn't hurt anymore after searching for the appropriate event and counter and wrote the metric formulas.

PR #244, opened 2019-10-22
Allow arbitrary thread numbers per work group in likwid-bench

ohlmann:
Up to now, different work groups must have the same thread number for likwid-bench to work properly. With this change, different work groups can have different numbers of threads, which is useful for e.g., when one wants to have 2 threads on the first socket and only one on the second.

Example: `likwid-bench -t copy -w S0:10MB:2 -w S1:10MB:1`

### DIFF
--- a/bench/includes/threads.h
+++ b/bench/includes/threads.h
@@ -32,6 +32,7 @@
 
 #include <pthread.h>
 #include <threads_types.h>
+#include <strUtil.h>
 
 #define THREADS_BARRIER pthread_barrier_wait(&threads_barrier)
 #define MIN_ITERATIONS 10
@@ -107,7 +108,8 @@ extern void threads_destroy(int numberOfGroups, int numberOfStreams);
 /**
  * @brief  Create Thread groups
  * @param  numberOfGroups The number of groups to create
+ * @param  groups Pointer to the groups data
  */
-extern void threads_createGroups(int numberOfGroups);
+extern void threads_createGroups(int numberOfGroups, Workgroup *groups);
 
 #endif /* THREADS_H */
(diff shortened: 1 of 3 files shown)


ohlmann:
I just saw that this probably resolves #210. My motivation was similar to what is mentioned there.

TomTheBear:
Thanks for the PR. I can understand the need for this patch but I do not recommend to use different number of threads per workgroup. If you have one group with X threads that saturate the memory bandwidth and one group with Y (Y << X), the second group probably executes the benchmark with higher performance as there is less competition/pressure on the memory interface.

Moreover, the patch is not complete. Until now, the output just takes the max time of all threads and the size of group0. The size and iterations per thread are taken just from group0 but should be presented separately for each group with different thread counts.

The #210 is not affected by this PR. For #210, the initialization of the arrays needs to be delayed until the threads are started (https://github.com/RRZE-HPC/likwid/blob/master/bench/likwid-bench.c#L506).

ohlmann:
Regarding the first point: I agree that there will be different performance in different groups, but one can change the size of the data (i.e. the workload) accordingly. Moreover, one can also produce load imbalances right now by using different sizes for different groups.

Regarding the second point: as far as I understand the code, `iters_per_thread` should be the same for all threads. With this, the output of "MFlops/s" and "MByte/s" should be fine because they use as size the sum over all threads. But already right now, the size per thread can be different for different groups because the size can be different, so I guess this would need to be adapted anyway?

In any case, I think one has to know what happens on the system in order to understand the numbers.

TomTheBear:
I'll merge the PR but add warnings for different sizes and different number of threads. Just to notice users that special attention is needed when using these features.

PR #257, opened 2019-12-06
Add temperature and voltage for Intel Core CPUs

jimmysitu:
Hi, All

I add two kinds of event here
1. Temperature of Intel Core processor PP0(CPU) and PP1(GPU)
2. Core voltage

Both event are tested on broadwell and skylake CPU

Thanks


TomTheBear:
Many thanks for the PR. Expect some delay (1-2 weeks) until I report back, I'm currently quite busy.

TomTheBear:
I checked your code and thanks for using similar coding style.

The PR consists of two features:

- CPU voltage readings
- Temperature readings for the Intel Desktop memory controllers

I found the CPU voltage registers and offsets in Intel SDM but I have **not** found the information for the temperature readings. Can you please point me to the documentation?

The CPU voltage readings are added to Intel Broadwell and Skylake. Why not the other architectures? According to SDM, it should be available since SandyBridge. Because you couldn't test it or are there other reasons?
Can you please make the shifting and masking in voltage.h file more precise? `result >> 32` is too coarse. SDM tells us, the value is 16 bit wide, so `(result >>32) & 0xFFFF` would be sufficient.

You have added 8 new types `MBOX0TMP` to `MBOX7TMP` but your code uses only `MBOX0TMP`. Is there a reason for this?

jimmysitu:
Hi, @TomTheBear

1. Temperature readings are mentioned in Intel CPUs [datasheet vol.2](https://www.intel.in/content/www/in/en/products/docs/processors/core/6th-gen-core-family-mobile-h-processor-lines-datasheet-vol-2.html). Section 7.77 and 7.78. Although it is first mentioned on a 6th gen core datasheet. I tested it works in 5th gen core.
2. I only have Broadwell and Skylake CPUs, and I not sure if this update works for others CPU family.
I think the voltage reading should works since Sandy bridge. 
3. MBOX0TMP to MBOX7TMP are copy cat as MBOX0 to MBOX7. Just for backup. I am not sure if there are so many MBOX on a multi socket server. 
4. I update the PR to mask the voltage more precise

TomTheBear:
Hi,

thanks for the link to the datasheet. There is quite some useful information in there besides the temperature readings. Why have you added only PP0 and PP1 but not the PKG or EDRAM domain (7.75, 7.76)? Can you please use a bitmask as well when reading the temperature, it's only the lower 8 bits, so please add a line `data &= 0xFF;` after reading the value from hardware. 

I'm fine with adding it for Skylake and Broadwell first. I have SandyBridge, Haswell and Coffeelake Desktop chips in my playground. I can test it after the merge of your PR.

The clientmem interface uses only MBOX0. The other types MBOX1-MBOX7 are only used for server-class chips which have 4 to 8 different PCI devices, one for each memory channel. The additional type MBOX0TMP is basically not needed, it just simplifies the calculation of the result. You can remove the additional types MBOX1TMP to MBOX7TMP.

The voltage part seems fine already and would be mergable but it's better to wait until the MBOXxTMP issues are resolved since it's a single PR.

I'm currently thinking about giving the clientmem interface a seperate type because it seems to work also on server-class chips (which use the MBOX0 type differently) because some of the information seems interesting for desktop and server chips (like the temperature readings).

jimmysitu:
Hi, @TomTheBear

1. Fix done
2. PKG temp seem can be read from MSR so I do not add it in MBOX0TMP
3. I am not sure EDRAM temp works since my CPUs are mobile version and do not have a EDRAM I suppose. Maybe you can try on your desktop

TomTheBear:
Looks good, Thanks.

TomTheBear:
I installed your PR to extend it to Intel Haswell desktop chips.

- Somehow the documentation of the voltage stuff is confusing. The bit field is `47:32` for the "Core Voltage" but the calculation contains only `37:32` (Table 2-20, Okt. 2019 version). I asked Intel about this.
- In your whole PR, I don't find the conversion from raw register value to actual core voltage value (value * 1/(2^13)). It would be best to have a separate function in voltage.h and use that in `calculateResult()` and `calculateMarkerResult()`, similar to the `power_getEnergyUnit()` function for RAPL readings.
- What is the meaning of `VTG` for the Core Voltage counter? Just for interest! I would have used something like `CVC` for Core Voltage Counters.

Just for info: I'm out of office until February, so don't expect any answers in January.

jimmysitu:
Hi, @TomTheBear 
- I update the conversion from raw register value to actual core voltage value,
but I involvoed a warning here. I am not sure how to deal with it in a better way.

./src/libperfctr.c:139:18: warning: implicit declaration of function ‘voltage_value’ [-Wimplicit-function-declaration]
         result = voltage_value(stop);
                  ^~~~~~~~~~~~~

- VTG is short for voltage. I think core voltage counter may not so good, since this is more like a sensor or meter.

TomTheBear:
I got a response from Intel, the bitfield for the voltage readings is `47:32`. So, no changes required.

To remove the warning, include the `voltage.h` file in `libperfctr.c`.

TomTheBear:
The addition of `voltage.h` to `libperfctr.c` does not work -> multiple references. 

I tested the voltage readings in MarkerAPI mode and it always returned 0. This is because the reference to `voltage_value()` is unresolved. I tested it by moving the calculation into `libperfctr.c`. This works. For the power module, there is `power.c` and `power.h`, so it might be beneficial to move the functions in a file `voltage.c` and just export the functions through `voltage.h`.

jimmysitu:
HI, @TomTheBear 

Fixed done. Thanks.

TomTheBear:
In the last commit, I added voltage support for all architectures after Intel SandyBridge. I havn't added the temperature readings. We could add it for all Desktop chips but for server-class chips we have to re-work the mmap interface. At the moment you have either mmap (desktop) or PCIe-based devices (server). The mmap interface and PCI devices share the `MBOX0` type, so we would need an extra type for the mmap interface.

PR #266, opened 2020-02-02
likwid-bench: improve peakflops micro benchmark

melven:
* add different variants AVX/AVX-512 with/without FMAs, similar to the other benchmarks
* higher unrolling of AVX/SSE variants to obtain better results (on Haswell)
* add missing description and annotation (number of instructions etc)

### DIFF
--- a/bench/x86-64/peakflops.ptt
+++ b/bench/x86-64/peakflops.ptt
@@ -2,6 +2,12 @@ STREAMS 1
 TYPE DOUBLE
 FLOPS 8
 BYTES 8
+DESC Double-precision multiplications and additions with a single load, only scalar operations
+LOADS 1
+STORES 0
+INSTR_CONST 18
+INSTR_LOOP 12
+UOPS 11
 movaps FPR1, [rip+SCALAR]
 movaps FPR2, [rip+SCALAR]
 LOOP 1
--- a/bench/x86-64/peakflops.ptt
+++ b/bench/x86-64/peakflops.ptt
@@ -1,24 +1,45 @@
 STREAMS 1
 TYPE DOUBLE
-FLOPS 8
+FLOPS 16
 BYTES 8
 DESC Double-precision multiplications and additions with a single load, only scalar operations
 LOADS 1
 STORES 0
-INSTR_CONST 18
-INSTR_LOOP 12
-UOPS 11
+INSTR_CONST 32
+INSTR_LOOP 20
+UOPS 19
 movaps FPR1, [rip+SCALAR]
 movaps FPR2, [rip+SCALAR]
+movaps FPR3, [rip+SCALAR]
+movaps FPR4, [rip+SCALAR]
+movaps FPR5, [rip+SCALAR]
+movaps FPR6, [rip+SCALAR]
+movaps FPR7, [rip+SCALAR]
+movaps FPR8, [rip+SCALAR]
+movaps FPR9, [rip+SCALAR]
+movaps FPR10, [rip+SCALAR]
+movaps FPR11, [rip+SCALAR]
+movaps FPR12, [rip+SCALAR]
+movaps FPR13, [rip+SCALAR]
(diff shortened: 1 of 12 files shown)


melven:
I know this is not a very useful benchmark at all. But it gives an idea of the peak performance.

Of course, one can look up / calculate the peak flops but it is a bit tedious with AVX frequencies and one or two FMAs per core depending on the CPU model.

I've tested the different variants on a Haswell and on a Skylake Gold node (with AVX512) and got quite useful results (with a data size that fits into L1).

I hope that the number of instructions and uOps are correct, these seem to include some constant part from the generated code.

Background:
My goal is to write a small script that obtains all information for some simple (roofline based) performance models...

TomTheBear:
Thanks for the PR. You are right that the peakflops can be calculated from the data sheet values and some inputs are system/runtime dependent (FMA units and CPU clock).

The peakflops benchmark was committed accidentally by me, it was just a test from my side and somehow slipped in. My main problem with this type of benchmark in likwid-bench is that it basically does not need to have a stream, it can run completely in registers. likwid-bench's original purpose are streaming-like kernels. That's why the original peakflops benchmark uses a single stream (and traverses it). That's why you need to set a stream size that fits into L1.

I started already to add support for in-register benchmarks but there are still some open questions like how to specify the workgroup (domain:size:threads) if you don't need a size.

I'll take a look in the next days.

TomTheBear:
Could you add SP variants as well? I can also do that after merging your PR.

melven:
I also added the corresponding single-precision variants.

Results on Haswell and Skylake seemed fine to me but please have a look if the code is correct (number of instructions/flops/bytes/loop counter etc). I'm not very used to programming assembler instructions...

TomTheBear:
The scalar versions (`peakflops` and `peakflops_sp`) are quite slow. Have you tested different register combinations to achieve higher flop rates?

melven:
No I didn't test different instruction/register combinations there.

On Haswell, I obtained quite accurately 2 Flops/Cycle independent of single/double-precision (as I expected).

On Skylake, I didn't really look closely enough for scalar operations... 2.3 Flops/Cycle on a Xeon Silver and 2.85 Flops/Cycle on a Xeon Gold. Should be 2 and 3 when I interpret the diagrams on wikichip correctly.

TomTheBear:
I played around a bit (Skylake Gold 6148, single core, 15kB vector length)

|benchmark| MFlops/s | scalars_per_vec | ops |ratio|
|--------|------|-----|-----|---|
|`peakflops`| 4786.83 | 1 | 1 |1 |
|`peakflops_sse` | 9561.76 | 2 | 1| 1.995 |
|`peakflops_axv`| 18991.50 | 4 | 1 | 3.963 |
|`peakflops_avx_fma` | 37972.02 | 4 | 2 | 7.924 |
|`peakflops_axv512` | 37890.97 | 8 | 1 | 7.907 |
|`peakflops_axv512_fma` | 75789.48 | 8 | 2 | 15.817 |

So, for scalar FP ops you get 1.995 FPops/cyc, for AVX  7.9 FPops/cyc and AVX512 15.8 FPops/cyc.

|benchmark| MFlops/s | scalars_per_vec | ops |ratio|
|--------|------|-----|-----|---|
|`peakflops_sp`| 4791.48 | 1 | 1 |1 |
|`peakflops_sp_sse` | 19117.01 | 4 | 1| 3.989 |
|`peakflops_sp_axv`| 37974.61 | 8 | 1 | 7.925 |
|`peakflops_sp_avx_fma` | 75954.47 | 8 | 2 |15.852 |
|`peakflops_sp_axv512` | 75540.67 | 16 | 1 | 15.766 |
|`peakflops_sp_axv512_fma` | 151602.06 | 16 | 2 |  31.640 |

So, with my changes to `peakflops` and `peakflops_sp` we are ready to merge. Thanks again for providing the benchmarks.

PR #296, opened 2020-06-16
Adding support for Fujitsu A64FX

TomTheBear:


TomTheBear:
Works well on Fujitsu FX700 but not on FX1000 due to the extra unusable cores and the strange system topology.

PR #299, opened 2020-06-23
Create virtual numa node in case detection fails

### LINKED ISSUE #290: likwid-topology segfault in affinity_finalize on Raspberry Pi 4 (ARMv8)
# Problem

Using a checkout of 99b0d23927f5e65cfa4eb5aeac1c57504395694b (also happens with the 5.0.1 release) compiled for a Raspberry Pi 4 causes both likwid-topology and likwid-pin (as far as I can tell due to the same issue) to segfault after execution.

As far as I can tell this is probably a similar problem to #228.

# Debugging

Debugging the code reveals that there is an out of bounds write to the `domains` array in `affinity_init`, which corrupts `domains[0].tag` (`{mlen = 18, slen = 1, data = 0x55555b6f60 "N"}` -> `{mlen = 18, slen = 1, data = 0x55555b6f60 "@p[UU"}`) and one way or another causes the call to `free` done at the end to crash.

The debug output shows that it seems to think 2 domains will be created, but actually 3 end up being made:
```
DEBUG - [affinity_init:304] Affinity: Socket domains 1
DEBUG - [affinity_init:312] Affinity: NUMA domains 0
DEBUG - [affinity_init:315] Affinity: CPUs per socket 4
DEBUG - [affinity_init:340] Affinity: All domains 2
DEBUG - [affinity_init:353] Affinity domain N: 4 HW threads on 4 cores
DEBUG - [affinity_init:392] Affinity domain S0: 4 HW threads on 4 cores
DEBUG - [affinity_init:491] Affinity domain M0: 4 HW threads on 4 cores
```

Running `likwid-pin -p` shows this corrupted string:
```
Domain  �:�U:
        0,1,2,3

Domain S0:
        0,1,2,3

[1]    6056 segmentation fault (core dumped)  likwid-pin -p
```

# Details

 * likwid Version: 99b0d23927f5e65cfa4eb5aeac1c57504395694b
 * Linux Kernel: Linux tobias-pi 5.4.38-1-ARCH #1 SMP PREEMPT Wed May 6 11:05:57 MDT 2020 aarch64 GNU/Linux
 * CPU Model: Raspberry Pi 4 Model B Rev 1.1
 * [cpuinfo.txt](https://github.com/RRZE-HPC/likwid/files/4658128/cpuinfo.txt)
 * [gdb.txt](https://github.com/RRZE-HPC/likwid/files/4658093/gdb.txt)
 * [likwid-topology.txt](https://github.com/RRZE-HPC/likwid/files/4658096/likwid-topology.txt)
 * [lscpu.txt](https://github.com/RRZE-HPC/likwid/files/4658098/lscpu.txt)

Tobi29:
This modifies the error handling to use a "virtual" NUMA node in case no proper one was found, in order to guarantee that there is always at least one, even on systems without any NUMA support (like on the Raspberry Pi 4).

Previously the hwloc backend already handled this on its own, so I extracted that code into a virtual "backend", which is used by both the proc and empty backends as well, in case no better info can be detected.

For memory usage the procfs is used, which was done by the hwloc backend as well, and for the  `numberOfProcessors` the `cpuid_topology.numHWThreads`  value is taken.

In order to avoid any behaviour changes the hwloc backend recomputes the `numberOfProcessors` on its own, not sure if that is a good or bad idea.

I have not gotten around to testing this on any machines with actual NUMA support, but the changes only cover previous error cases that should never occur with multiple NUMA nodes.

I added myself as an author in the two files I created, if it would be preferred without that, that would be fine as well. (The code is mostly pasted from existing parts anyways)

Comments on issues or better approaches to this would be very welcome :)

Fixes #290, showing sensible data there now (CPU type is still missing, but that should be unrelated):
```
--------------------------------------------------------------------------------
CPU name:       BCM2835
CPU type:       nil
CPU stepping:   3
********************************************************************************
Hardware Thread Topology
********************************************************************************
Sockets:                1
Cores per socket:       4
Threads per core:       1
--------------------------------------------------------------------------------
HWThread        Thread          Core            Socket          Available
0               0               0               0               *
1               0               1               0               *
2               0               2               0               *
3               0               3               0               *
--------------------------------------------------------------------------------
Socket 0:               ( 0 1 2 3 )
--------------------------------------------------------------------------------
********************************************************************************
Cache Topology
********************************************************************************
********************************************************************************
NUMA Topology
********************************************************************************
NUMA domains:           1
--------------------------------------------------------------------------------
Domain:                 0
Processors:             ( 0 1 2 3 )
Distances:              10
Free memory:            1570.78 MB
Total memory:           1865.91 MB
--------------------------------------------------------------------------------
```

### DIFF
--- a/src/numa_hwloc.c
+++ b/src/numa_hwloc.c
@@ -262,36 +262,13 @@ hwloc_numa_init(void)
 #else
         hwloc_type = HWLOC_OBJ_SOCKET;
 #endif
-        numa_info.numberOfNodes = 1;
-
-        numa_info.nodes = (NumaNode*) malloc(sizeof(NumaNode));
-        if (!numa_info.nodes)
+        if (virtual_numa_init() < 0)
         {
-            fprintf(stderr,"No memory to allocate %ld byte for nodes array\n",sizeof(NumaNode));
             return -1;
         }
-        numa_info.nodes[0].id = 0;
+
+        /* Use hwloc to set numberOfProcessors */
         numa_info.nodes[0].numberOfProcessors = 0;
-        numa_info.nodes[0].totalMemory = getTotalNodeMem(0);
-        numa_info.nodes[0].freeMemory = getFreeNodeMem(0);
-        numa_info.nodes[0].processors = (uint32_t*) malloc(numPUs * sizeof(uint32_t));
-        if (!numa_info.nodes[0].processors)
-        {
-            fprintf(stderr,"No memory to allocate %ld byte for processors array of NUMA node %d\n",
-                    numPUs * sizeof(uint32_t),0);
-            return -1;
-        }
-        numa_info.nodes[0].distances = (uint32_t*) malloc(sizeof(uint32_t));
-        if (!numa_info.nodes[0].distances)
-        {
-            fprintf(stderr,"No memory to allocate %ld byte for distances array of NUMA node %d\n",
-                    sizeof(uint32_t),0);
-            return -1;
-        }
-        numa_info.nodes[0].distances[0] = 10;
-        numa_info.nodes[0].numberOfDistances = 1;
-        cores_per_socket = cpuid_topology.numHWThreads/cpuid_topology.numSockets;
-
         for (d=0; d<likwid_hwloc_get_nbobjs_by_type(hwloc_topology, hwloc_type); d++)
         {
             obj = likwid_hwloc_get_obj_by_type(hwloc_topology, hwloc_type, d);
--- a/src/numa_hwloc.c
+++ b/src/numa_hwloc.c
@@ -253,30 +253,7 @@ hwloc_numa_init(void)
        aggregate all sockets in the system into the single virtually created NUMA node */
     if (numa_info.numberOfNodes == 0)
     {
-#if defined(__ARM_ARCH_7A__) || defined(__ARM_ARCH_8A__)
-#if HWLOC_API_VERSION > 0x00020000
-        hwloc_type = HWLOC_OBJ_NUMANODE;
-#else
-        hwloc_type = HWLOC_OBJ_NODE;
-#endif
-#else
-        hwloc_type = HWLOC_OBJ_SOCKET;
-#endif
-        if (virtual_numa_init() < 0)
-        {
-            return -1;
-        }
-
-        /* Use hwloc to set numberOfProcessors */
-        numa_info.nodes[0].numberOfProcessors = 0;
-        for (d=0; d<likwid_hwloc_get_nbobjs_by_type(hwloc_topology, hwloc_type); d++)
-        {
-            obj = likwid_hwloc_get_obj_by_type(hwloc_topology, hwloc_type, d);
-            /* depth is here used as index in the processors array */
-            depth = d * cores_per_socket;
-            numa_info.nodes[0].numberOfProcessors += likwid_hwloc_record_objs_of_type_below_obj(
-                    likwid_hwloc_topology, obj, HWLOC_OBJ_PU, &depth, &numa_info.nodes[0].processors);
-        }
+        return virtual_numa_init();
     }
     else
     {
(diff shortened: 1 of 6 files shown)


TomTheBear:
Thanks for the PR. I'm not 100% sure whether this extra code is needed or we just have to sanitize the NUMA nodes values to 1 somewhere in the hwloc backend. Also the copy of `getFreeNodeMem` and `getTotalNodeMem` is probably not required and could be just use the function from the hwloc file.

Of course you are the author, we would never change that.

TomTheBear:
I havn't forgot your PR. I logged in to a RasPi4 but it was installed with a 32 Bit Raspbian and there was no chance to test your code. Since it's remote access, I have to see how I can install a more recent OS.

TomTheBear:
I tried today on a RasPi4 and as  I thought, a sanitation to 1 is basically enough:
`if (numberOfNumaDomains < 1) { numberOfNumaDomains = 1; }` in `src/affiniy.c`
But `likwid-topology` won't print the NUMA domain because `numa_info.numberOfNodes` is still zero. Other application that access the affinity domains (e.g. `likwid-pin`) work with the fix.

Since it is not a clean fix, I will check your virtual NUMA domain code again.

The detection of the micro architecture is no problem, the A72 is just missing in the big switch-case statements in `src/topology.c` and `src/perfmon.c`.

src/numa_virtual.c
```diff
@@ -0,0 +1,166 @@
...
+    numa_info.nodes[0].processors = (uint32_t*) malloc(cpuid_topology.numHWThreads * sizeof(uint32_t));
+    if (!numa_info.nodes[0].processors)
+    {
+        fprintf(stderr,"No memory to allocate %ld byte for processors array of NUMA node %d\n",
+                cpuid_topology.numHWThreads * sizeof(uint32_t),0);
+        return -1;
+    }
+    for (i = 0; i < cpuid_topology.numHWThreads; i++)
+    {
+        numa_info.nodes[0].processors[i] = i;
```
TomTheBear, line 153: Please use `cpuid_topology.threadPool[i].apicId`, just `i` might not be right from strangely numbered hw threads.

src/numa_virtual.c
```diff
@@ -0,0 +1,166 @@
...
+    }
+    numa_info.nodes[0].distances = (uint32_t*) malloc(sizeof(uint32_t));
```
TomTheBear, line 155: Do all allocations in the beginning of the function before setting any values, so move this `malloc` some lines up. Make sure to `free` allocated memory when returning early (your current code does not free `numa_info.nodes[0].processors` if allocation for `numa_info.nodes[0].distances` fails)

src/numa_virtual.c
```diff
@@ -0,0 +1,166 @@
...
+
+#include <error.h>
+
+#include <numa.h>
+#include <topology.h>
+
+/* #####   FUNCTION DEFINITIONS  -  LOCAL TO THIS SOURCE FILE   ########### */
+
+static uint64_t
+getFreeNodeMem(void)
```
TomTheBear, line 45: Put the function in `src/numa.c`, export it via `src/includes/numa.h` and use it here and in the other `numa_*` files.

src/numa_virtual.c
```diff
@@ -0,0 +1,166 @@
...
+    {
+        bdestroy(freeString);
+        ERROR;
+    }
+    bdestroy(freeString);
+    return free;
+}
+
+static uint64_t
+getTotalNodeMem(void)
```
TomTheBear, line 85: Put the function in `src/numa.c`, export it via `src/includes/numa.h` and use it here and in the other `numa_*` files.

src/numa_hwloc.c
```diff
@@ -262,36 +262,13 @@ hwloc_numa_init(void)
...
-        {
-            fprintf(stderr,"No memory to allocate %ld byte for distances array of NUMA node %d\n",
-                    sizeof(uint32_t),0);
-            return -1;
-        }
-        numa_info.nodes[0].distances[0] = 10;
-        numa_info.nodes[0].numberOfDistances = 1;
-        cores_per_socket = cpuid_topology.numHWThreads/cpuid_topology.numSockets;
 
+        /* Use hwloc to set numberOfProcessors */
```
TomTheBear, line 270: Wouldn't this part overwrite the settings you have done in `virtual_numa_init()`?

Issue #297, opened 2020-06-18
Improve documentation


rileyweber13:
Similar to issue #273. Issue #201 is also relevant.

I feel like the documentation could be improved, and I would like to help. Some things I would have liked to have in the documentation as I started learning to use LIKWID:

- [ ] An up front explanation of the differences between the marker API and perfmon API and their use-cases. 
- [ ] Information about how to avoid using the CLI if one wishes to extend likwid or use it in a larger project.
- [ ] Working example code for the internal marker api. The file at `examples/C-internalMarkerAPI.c` does not work with the current version of likwid.

Are these good things to start with? Or are there other priorities for the documentation?

I noticed the documentation is spread across the wiki, doxygen comments, man pages, and the readme file. If I were to begin improving the documentation, what medium should I use? If doxygen comments, the readme, or man pages are preferred, I can make a pull request. But if the wiki is preferred, how can I ensure changes are approved?

I'd also be happy to work on creating a static site (as mentioned in #273) that could be maintained with pull requests and hosted on github pages, if the likwid team would like to take the documentation in that direction. This could be easily done with the man pages.

rileyweber13:
I discovered there's already a [static site displaying the doxygen docs](https://rrze-hpc.github.io/likwid/Doxygen/). Improvements to the documentation should add to what's already there. Perhaps the wiki and readme should make it easier to find these docs?

TomTheBear:
Please check the [MarkerAPI pitfalls page](https://github.com/RRZE-HPC/likwid/wiki/LikwidMarkerAPIPitfalls) whether anything is missing. Just add bullet points. I have the developer glasses on and probably won't cover everything that might interest users.

TomTheBear:
For your first bullet point: https://github.com/RRZE-HPC/likwid/wiki/LikwidAPI-and-MarkerAPI

rileyweber13:
Hey Thomas! I appreciate you working on this, it will certainly help me (and others, I hope) work with likwid. I'll take a look at your additions and let you know any thoughts I have.

TomTheBear:
You have mentioned static web page in your opening post. If you have some experiences with it and possibly Github Pages, we could create a better page with Links, API documentation per version and so on. I just had the idea because I wanted to update the API documentation. The current version does not list the Nvidia GPU support. Would you be interested in doing that? In a second step we could set up a Github Action that adds the documentation for released versions automatically.

rileyweber13:
[...]

Issue #292, opened 2020-05-28
Unreasonably high values (not deterministic)


TomTheBear:
See https://groups.google.com/forum/#!topic/likwid-users/m1ElsBTerfk

Thanks @rileyweber13 for reporting

rileyweber13:
Hey Thomas!

In case we have more problems with the google group and for the convenience of people reading this for the first time, I've included links to the code I'm using to demonstrate this issue. I use the c code [likwid_minimal.c](https://github.com/rileyweber13/fine-hardware-visualization/blob/5c0306d2bdd6b5735765bd3739b34df9d123fbbc/tests/likwid_minimal.c) and the bash script [likwid_minimal_repeated.sh](https://github.com/rileyweber13/fine-hardware-visualization/blob/5c0306d2bdd6b5735765bd3739b34df9d123fbbc/tests/likwid_minimal_repeated.sh) to count the number of failures in a given number of tests.

Anyways, I realized while working on this issue that I had some leftover files in `/usr/local` that I think were getting picked up at runtime, and I think that was muddying the water. I removed those, ran `make clean` and `make distclean`, and then installed likwid to `/usr/local/likwid-master`. I did the same with version 4.3.4, but installed it to `/usr/local/likwid-v4.3.4/`. I then tested each version with a single group and with multiple groups. Each case was tested both with hyperthreading enabled and with hyperthreading disabled. Since my machine does not provide the option to disable hyperthreading in the bios, I disabled it by using LIKWID_THREADS to select cores 0,1 (which represent full cores and not hyperthreads in my case) and then set the number of openMP threads to be 2. Here are my results:
 - Version 4.3.4. Compiled with `g++ likwid_minimal.c -L/usr/local/likwid-v4.3.4/lib -I/usr/local/likwid-v4.3.4/include -llikwid -mtune=native -fopenmp -o likwid_minimal` and ran with `LD_LIBRARY_PATH=/usr/local/likwid-v4.3.4/lib PATH=/usr/local/likwid-v4.3.4/sbin:$PATH ./likwid_minimal_repeated.sh`
   - No hyperthreading, specifying one group (L2): 0 failures out of 100 tests
   - With hyperthreading, specifying one group (L2): 0 failures out of 200 tests
     - there was one case where I received the error "WARN: Skipping region (null) for evaluation"
   - No hyperthreading, specifying multiple groups (L2|L3|FLOPS_SP|FLOPS_DP): 3 failures out of 200 tests
[...]

PR #303, opened 2020-07-06
#297 improve examples

rileyweber13:
`C-markerAPI` is a simple example and `C-internalMarkerAPI` is more complete. Examples now work with current version of likwid (v5.0.1) and comments have been dramatically improved.

Mentioned in RRZE-HPC/likwid#292
Part of RRZE-HPC/likwid#297

### DIFF
--- a/examples/C-internalMarkerAPI.c
+++ b/examples/C-internalMarkerAPI.c
@@ -96,7 +96,7 @@ int main(int argc, char* argv[])
      * for all architectures supported by likwid except nvidiagpus and Xeon Phi
      * (KNC). 
      */
-    setenv("LIKWID_EVENTS", "L2|FLOPS_DP|INSTR_RETIRED_ANY:FIXC0", 1);
+    setenv("LIKWID_EVENTS", "FLOPS_DP|L2|INSTR_RETIRED_ANY:FIXC0", 1);
     // setenv("LIKWID_EVENTS", "BRANCH|INSTR_RETIRED_ANY:FIXC0", 1);
 
     /* LIKWID_THREADS must be set to the list of hardware threads that we will
@@ -192,6 +192,11 @@ int main(int argc, char* argv[])
 
 #pragma omp parallel
 {
+    /* first, we will demonstrate measuring a single region, getting results
+     * with LIKWID_MARKER_GET, and resetting the region so that these results
+     * do not affect later measurements
+    */
+    
     /* variables needed for LIKWID_MARKER_GET */
     int nr_events = MAX_NUM_EVENTS;
     double events[MAX_NUM_EVENTS];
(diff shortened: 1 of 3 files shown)


TomTheBear:
Hi Riley. Thanks for the PR. The code looks well in general and is much more complete.

Some comments  for `C-internalMarkerAPI.c`:
* Maybe we shouldn't register one of the regions to show that it is optional?
* unsigned data types (`ull`) shouldn't be used as iterator variable (sometimes prevents vectorization)
* Are you sure about the `omp barriers` in the loop? In my understanding of the MarkerAPI, they are not required. Only for some special cases like `LIKWID_MARKER_SWITCH` because they need to be executed in serial although being close to or in parallel regions (guarded with omp single/master).

rileyweber13:
Hey Thomas! I'll change it so one of the regions isn't registered and change the datatype to signed. Thanks for the tips.

As for the omp barriers, without them I get the unreasonably high values discussed in #292 . I suppose the barriers are more of a workaround than a fix, but in my case they were absolutely necessary because without them about 10% of the machine's cores were giving garbage data. (Or, in the case of a development machine with 4 cores, one core would fail every 5-6 trials). That being said, I think it would be good to verify this, can you test it on your machine? If so, would you like for me to throw together a test case? Or is the example code discussed in #292 sufficient?

Also, about `LIKWID_MARKER_SWITCH`: as far as I know, `#pragma omp single` only synchronizes at the end of the block and if you try to switch before all threads have stopped then you get errors. Additionally, `#pragma omp critical` performs no synchronization. Therefore, the barrier is still necessary if you want to execute it in a `single` or `critical` block nested within a parallel region.

rileyweber13:
Thomas, 

I experimented a bit further today and I think I was wrong, actually. The file I used to demonstrate the unreasonably high values was `likwid_minimal.c`. Before I started using the `LIKWID_MARKER_NEXT() = likwid_markerNextGroup()` in a sequential region, the main loop of my code was the following:

```c
#pragma omp parallel
  {
    for (int j = 0; j < 8; j++)
    {
#pragma omp barrier
      likwid_markerStartRegion("double_flops");
      do_flops(a, b, c, NUM_FLOPS);
      likwid_markerStopRegion("double_flops");
#pragma omp barrier
      likwid_markerStartRegion("copy");
      do_copy(arr, copy_arr, n, NUM_COPIES);
      likwid_markerStopRegion("copy");
#pragma omp barrier
      likwid_markerNextGroup();
    }
  }
```

Removing either the first or last barrier caused lots of problems with either `WARN: Region <region> already stated`, `WARN: Stopping an unknown/not-started region <region>`, or unreasonably high values. I'm not sure why I had the middle barrier, but I think it was leftover from earlier tests. Perhaps I had placed it before I had a barrier preceeding `likwid_markerNextGroup()` and it alleviated some of the problems, but it became unnecessary after adding the barrier before `likwid_markerNextGroup()`.

Once you told me that group switching should take place in a serial region, I changed the code to the following:

```c
#pragma omp parallel
  {
    for (int j = 0; j < 8; j++) 
    {
      likwid_markerStartRegion("double_flops");
      do_flops(a, b, c, NUM_FLOPS);
      likwid_markerStopRegion("double_flops");
#pragma omp barrier
      likwid_markerStartRegion("copy");
      do_copy(arr, copy_arr, n, NUM_COPIES);
      likwid_markerStopRegion("copy");

      #pragma omp barrier
      #pragma omp single
      {
        likwid_markerNextGroup();
      }
    }
  }
```

Just now I tested this without the barrier in the middle and didn't get any problems on my test machine after 100 iterations. I thought I had tested this already, hence my confidence from earlier, but I guess I was wrong.

I'm worried that this result is not conclusive, and that the barrier may still be needed at times. What if I remove the barriers in the example code, but make a note in the comments and in the docs that if multiple regions are in one parallel block, it may sometimes be helpful to have a barrier?

TomTheBear:
Hi Riley,

you are right about `SWITCH`, `omp single` (and the others) don't have an entry barrier. So, this barrier is definitely needed.

The barrier in between is not required because threads are allowed to execute the regions independently and can even be in a completely different iteration. That's how we planned it and (hopefully) implemented it.

How about we make the examples clean and simple but put a link to the wiki/README/APIdocs in there pointing to a page with possible pitfalls including short code snippets and explaination? I think users are searching for info in the wiki/README/APIdocs and not that much in code. Of course, that doesn't mean that we cannot document it in the example as well but we have multiple locations for the same information and have to update it accordingly in case of changes.

My main problem is to tell users that "sometimes something is needed" without being clear about the acutal situations. Multiple regions in a block are not the problem, multiple regions in block in combination with `LIKWID_MARKER_SWITCH` (or any other call that should be executed serially) _can_ be a problem. I don't want users to put unneeded barriers in their code to be sure that LIKWID works _always_.

rileyweber13:
I think the reason that I put "sometimes you need this" is that I've had so many stability issues with likwid that I'm not confident using the barrier with `LIKWID_MARKER_NEXT` and running it only by one thread are sufficient to prevent the problems I've experienced. However, since you have a much deeper understanding of the code, I'll defer to your judgment. I'll remove the comments about barriers, but I do think having a page for common pitfalls would be good. That being said, I don't feel well-equipped to write that page. 

In general, would you like me to reduce the verbosity of comments and move that information to somewhere else? If so, is the README or doxygen pages a better place for that information?

TomTheBear:
Hi Riley,

sorry for my late comments after you did all that work. We keep it like it is including your comment about the barriers. As soon as I write the pitfalls page, we add a link in the comments as well, so it's easy to get deeper information when it's needed.

Your suggestion to have a separate loop for GET/RESET seems a good idea. That would clean the code up and we would have one loop with the basic macros and one with all capabilities.

rileyweber13:
Okay, for now I'll keep the comments and just simplify the calls to GET/RESET. 

No worries about the comments, I think it's important to get the example right since it conveys so much information about the intended usage. Also, I agreee that the documentation should be improved and I'd like to help with that.

I'll ping you again once I make the changes to the PR.

TomTheBear:
I started with a pitfalls page: https://github.com/RRZE-HPC/likwid/wiki/LikwidMarkerAPIPitfalls
It is not included in the menu yet, I'll do that when I have covered all aspects.

rileyweber13:
The pitfall page has some valuable information, thanks. Will you also mention something about measuring multiple groups across the same region? It's kind of selfish for me to ask, but it's an important use-case for me :grin: 

Also, I made a separate section for GET/RESET. I think this makes the example clearer. Is there anything else you'd like me to change?

TomTheBear:
Sure, I will add more and more info on the page. I wanted to start with basic stuff that most users need.

If you have more ideas or missing documentation, we could add a Todo page to the wiki. 

The code looks good like that. I'll check it on the weekend.

rileyweber13:
I think a todo wiki page would be good. The biggest things I would like to see are: 1) an explanation of the difference between the marker API and perfmon API and when to use each and 2) more complete function tips, like when to use each function in a parallel or serial region. 

A "todo" page would also help me know how I can contribute to the documentation

TomTheBear:
Hi Riley, great examples.

I have two more minor comments:

- Can you put a check in `C-internalMarkerAPI.c` when you run it wrapped by `likwid-perfctr`, so the environment variables are already set. Just print out an Error noting that this example should be run without `likwid-perfctr` and exit. If you run it with `likwid-perfctr` alone, you get numbers but they are not for the whole run. If you run with `likwid-perfctr -m` it complains in the end about the missing MarkerAPI file. You can only check the last case with certainty because `likwid-perfctr` alone does not always set an enviroment variable. My recommendation would be env variable `LIKWID_PIN` as `likwid-perfctr` is commonly run with pinning enabled.
- Please add yourself to the header of `C-markerAPI.c`

rileyweber13:
I added a check, which I hope is sufficient. It checks if `LIKWID_PIN` is set, prints an error message, and then exits. Sometimes likwid-perfctr still presents counts, but I am not sure how to stop this.

I also added my name. 

Let me know if there's anything else you need from me!

TomTheBear:
It's fine that likwid-perfctr prints numbers in some cases as long as there is the message that it doesn't work. We cannot catch all cases.
I think we are ready to merge.

PR #309, opened 2020-07-30
Updates likwid-config.cmake to support find_package

jrmadsen:
- added test/cmake for testing whether the find_package works
- modified installation location (required for find_package to work
- Validation test:

```console
# build and install likwid
make && make install
# go to cmake test folder
cd test/cmake
# make build directory
mkdir build && cd build
# configure where PREFIX is the entry in config.mk
CMAKE_PREFIX_PATH=${PREFIX} cmake ..
# build the test executable
make
```

I had to change the installation location to get `find_package(likwid)` to work correctly, [here are the options for Unix installs](https://cmake.org/cmake/help/latest/command/find_package.html):

```console
<prefix>/(lib/<arch>|lib*|share)/cmake/<name>*/                 
<prefix>/(lib/<arch>|lib*|share)/<name>*/                       
<prefix>/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/         
<prefix>/<name>*/(lib/<arch>|lib*|share)/cmake/<name>*/         
<prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/               
<prefix>/<name>*/(lib/<arch>|lib*|share)/<name>*/(cmake|CMake)/ 
```


TomTheBear:
Thanks. I asked my colleague to look at it because I have absolutely no clue about cmake. From my side, the PR mergable.

You mean the install location for the cmake file? That's fine. When I first read your post I thought shortly you changed the whole folder structure of installed LIKWID. :sweat_smile:

test/cmake/CMakeLists.txt
```diff
@@ -0,0 +1,8 @@
+cmake_minimum_required(VERSION 3.11 FATAL_ERROR)
+
+project(Likwid-Test LANGUAGES C)
+
+find_package(likwid REQUIRED COMPONENTS marker OPTIONAL_COMPONENTS nvmarker)
+
```
christiealappatt, line 6: The test doesn't compile since it doesn't find include path, add `include_directories(${LIKWID_INCLUDE_DIR})`

Makefile
```diff
@@ -581,7 +592,9 @@ move: move_daemon move_freq move_appdaemon
 	@chmod 755 $(LIKWIDFILTERPATH)
 	@cp -f $(abspath $(PREFIX)/share/likwid/filter)/* $(LIKWIDFILTERPATH)
 	@chmod 755 $(LIKWIDFILTERPATH)/*
-	@install -m 644 $(LIBPREFIX)/likwid-config.cmake $(INSTALLED_LIBPREFIX)
+	@mkdir -p $(INSTALLED_LIBPREFIX)/cmake/likwid
+	@chmod 755 $(INSTALLED_LIBPREFIX)/cmake/likwid
+	@install -m 644 $(LIBPREFIX)/likwid-config.cmake $(INSTALLED_LIBPREFIX)/share/likwid
```
christiealappatt, line 597: shouldn't install be at `$(INSTALLED_LIBPREFIX)/cmake/likwid` instead of `$(INSTALLED_LIBPREFIX)/share/likwid`

jrmadsen: CMake has several paths it searches (that list I posted above) so either works. I chose the share path bc the directory is already made from other install commands

test/cmake/CMakeLists.txt
jrmadsen: Are you testing from the build tree? This test requires a make install

jrmadsen: Also this is modern CMake. You don't call include_directories, the include directory is automatically included when you link to "likwid::likwid" because I did `target_include_directories(likwid::likwid INTERFACE ${LIKWID_INCLUDE_DIRS})`

jrmadsen: If you follow the validation steps I laid out above, you'll see that it does indeed compile and the pre-processor flags are added

TomTheBear:
During my first flight over your PR I saw
```
cmake_minimum_required(VERSION 3.11 FATAL_ERROR)
```
but thought, it's probably some really old version. Turns out it isn't. I have Ubuntu 18.04 installed and the latest cmake that is provided is 3.10.2. Is there a way to achieve the same functionality with "legacy CMake"?

Makefile
TomTheBear: The problem that I see is that in the `install` target the file is installed to `$(PREFIX)/share/likwid/likwid-config.cmake` but in `move` you use `$(LIBPREFIX)/likwid-config.cmake=$(PREFIX)/lib/likwid-config.cmake` as source for the file. The `move` target should move an already `install`ed LIKWID into another location. So, the cmake file probably needs some sed magic again to contain the proper paths in the end.

jrmadsen: Well I corrected not using `$(INSTALLED_LIBPREFIX)` and `$(INSTALLED_BINPREFIX)` in addition to adding the sed magic (which appears to not be necessary now).

jrmadsen:
@TomTheBear Yea I forgot to change that. For whatever reason, my brain just cannot remember `cmake_minimum_required(...)`... I _always_ end up typing `cmake_required_version(...)` so I've developed a habit of just doing a copy + paste from whatever cmake project I have open at the time.

Theoretically, v3.3.2 is the minimum but I could only find a pre-built 3.5 so I went with that:

```console
$ cmake --version && cmake .. -G Ninja && ninja -v
cmake version 3.5.0

CMake suite maintained and supported by Kitware (kitware.com/cmake).
-- The C compiler identification is GNU 8.4.0
-- Check for working C compiler using: Ninja
-- Check for working C compiler using: Ninja -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Found likwid: /usr/local  found components:  marker nvmarker 
-- 
-- likwid_FOUND: 1
-- LIKWID_INCLUDES: /usr/local/include
-- LIKWID_LIBRARIES: /usr/local/lib/liblikwid.so
-- LIKWID_ROOT_DIR: /usr/local
-- LIKWID_NVIDIA_INTERFACE: true
-- LIKWID_FORTRAN_INTERFACE: false
-- LIKWID_INCLUDE_DIR: /usr/local/include
-- LIKWID_LIBRARY: /usr/local/lib/liblikwid.so
-- LIKWID_INCLUDE_DIRS: /usr/local/include
-- LIKWID_LIBRARIES: /usr/local/lib/liblikwid.so
-- LIKWID_LIBRARY_DIRS: /usr/local/lib
-- LIKWID_bench_EXECUTABLE: /usr/local/bin/likwid-bench
-- LIKWID_features_EXECUTABLE: /usr/local/bin/likwid-features
-- LIKWID_genTopoCfg_EXECUTABLE: /usr/local/bin/likwid-genTopoCfg
-- LIKWID_lua_EXECUTABLE: /usr/local/bin/likwid-lua
-- LIKWID_memsweeper_EXECUTABLE: /usr/local/bin/likwid-memsweeper
-- LIKWID_mpirun_EXECUTABLE: /usr/local/bin/likwid-mpirun
-- LIKWID_perfctr_EXECUTABLE: /usr/local/bin/likwid-perfctr
-- LIKWID_perfscope_EXECUTABLE: /usr/local/bin/likwid-perfscope
-- LIKWID_pin_EXECUTABLE: /usr/local/bin/likwid-pin
-- LIKWID_powermeter_EXECUTABLE: /usr/local/bin/likwid-powermeter
-- LIKWID_setFrequencies_EXECUTABLE: /usr/local/bin/likwid-setFrequencies
-- LIKWID_topology_EXECUTABLE: /usr/local/bin/likwid-topology
-- likwid_gotcha_FOUND: 
-- likwid_hwloc_FOUND: 
-- likwid_lua_FOUND: 
-- likwid_pin_FOUND: 
-- likwid_appDaemon_FOUND: 
-- likwid_marker_FOUND: ON
-- likwid_nvmarker_FOUND: ON
-- 
-- Configuring done
-- Generating done
-- Build files have been written to: /home/likwid/test/cmake/build
[1/2] /usr/bin/cc  -DLIKWID_NVMON -DLIKWID_PERFMON   -MMD -MT CMakeFiles/test_install_tree.dir/test.c.o -MF CMakeFiles/test_install_tree.dir/test.c.o.d -o CMakeFiles/test_install_tree.dir/test.c.o   -c ../test.c
[2/2] : && /usr/bin/cc     CMakeFiles/test_install_tree.dir/test.c.o  -o test_install_tree  /usr/local/lib/liblikwid.so -Wl,-rpath,/usr/local/lib && :
```

TomTheBear:
I just tried it and it works fine. Message is similar to yours and I get a warning that "LIKWID_NVMON is not defined" which is true for this installation.

Makefile
```diff
@@ -202,13 +202,23 @@ $(L_HELPER):
...
 
 $(DYNAMIC_TARGET_LIB): $(BUILD_DIR) $(PERFMONHEADERS) $(OBJ) $(TARGET_HWLOC_LIB) $(TARGET_LUA_LIB)
 	@echo "===>  CREATE SHARED LIB  $(TARGET_LIB)"
 	$(Q)${CC} $(DEBUG_FLAGS) $(SHARED_LFLAGS) -Wl,-soname,$(TARGET_LIB).$(VERSION).$(RELEASE) $(SHARED_CFLAGS) -o $(DYNAMIC_TARGET_LIB) $(OBJ) $(LIBS) $(TARGET_HWLOC_LIB) $(TARGET_LUA_LIB) $(RPATHS)
 	@ln -sf $(TARGET_LIB) $(TARGET_LIB).$(VERSION).$(RELEASE)
-	@sed -e s+'@PREFIX@'+$(INSTALLED_PREFIX)+g make/likwid-config.cmake > likwid-config.cmake
+	@sed -e s+'@PREFIX@'+$(INSTALLED_PREFIX)+g \
+		-e s+'@NVIDIA_INTERFACE@'+$(NVIDIA_INTERFACE)+g \
+		-e s+'@FORTRAN_INTERFACE@'+$(FORTRAN_INTERFACE)+g \
+		-e s+'@LIBPREFIX@'+$(LIBPREFIX)+g \
```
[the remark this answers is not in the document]
jrmadsen: Yep, good catch.

jrmadsen:
@christiealappatt

christiealappatt:
Sorry I forgot about that. PR has been merged.

Issue #317, opened 2020-08-20
Add support for Nvidia CUPTI Profiler API


TomTheBear:
The currently used CUPTI Events and Metrics API is deprecated and not usable on Nvidia GPUs with Compute Capability > 7.0. In order to support newer GPUs, we need the CUPTI Profiler API with PerfWorks for the metric definitions.

Issue #312, opened 2020-08-06
LIKWID NvMarkerAPI segfaults on Quadro RTX 5000


TomTheBear:
Hi,

I tried this with a simple CUDA kernel.

I have two Quadro RTX 5000 GPUs and use cuda/10.1.243.

It compiles fine with: 
`$ nvcc -DLIKWID_NVMON -L/home/bgursoy/likwid/5.0.1/lib -I/home/bgursoy/likwid/5.0.1/include matrixmult.cu -llikwid`

I run as follows:
```
$ likwid-perfctr -G 0 -W MEM -m ./a.out
--------------------------------------------------------------------------------
CPU name:	Intel(R) Xeon(R) CPU E5-2630 v3 @ 2.40GHz
CPU type:	Intel Xeon Haswell EN/EP/EX processor
CPU clock:	2.39 GHz
NVIDIA PerfWorks API current not supported. Trying CUPTI
Segmentation fault (core dumped)
```

It says here: https://developer.nvidia.com/perfworks that Quadra is a supported platform of PerfWorks however I am getting the warning. It looks like it also doesn’t work with CUPTI. (The code is correct.) 

What would be causing the issue?
Thanks for your time.

Regards,
Buket

_Originally posted by @buketbenek in https://github.com/RRZE-HPC/likwid/issues/311#issuecomment-669194791_

TomTheBear:
The message
`NVIDIA PerfWorks API current not supported. Trying CUPTI`
is not about the capabilities of your GPU but of LIKWID. LIKWID currently supports only event definitions coming from CUPTI but not from PerfWorks. The PerfWorks API is still under development and only partly published. At least the last time I have checked it.

I think the problem is that the function pointer is not properly set up due to the detection of a more recent GPU that supports PerfWorks. The fallback seems not to work. I'll check that.

buketbenek:
Dear Thomas,

Thanks a lot for your reply. Below is the same output for the provided example of likwid.

$ nvcc -I/home/bgursoy/likwid-5.0.1/include -DLIKWID_NVMON triad.cu -L/home/bgursoy/likwid-5.0.1/lib -llikwid
$  
$ ./a.out 
Running without GPU Marker API. Activate GPU Marker API with -m, -G and -W on commandline.
 11.05ms 388.76GB/s 
$ 
$ likwid-perfctr -G 0 -W MEM -m ./a.out
--------------------------------------------------------------------------------
[...]

PR #325, opened 2020-09-19
Nvmon Interface for Nvidia Profiling API (for Nvidia Volta and newer)

TomTheBear:
Works but not validated. Currently only tested with Cuda 10.1. Some events require to measure multiple passes to get a single result. That's currently not supported.


TomTheBear:
Related:
- https://github.com/RRZE-HPC/likwid/issues/317
- https://github.com/RRZE-HPC/likwid/issues/312

PR #349, opened 2020-12-10
docs: fix simple typo, sccess -> access

timgates42:
There is a small typo in bench/includes/likwid.h, src/includes/likwid.h.

Should read `access` rather than `sccess`.

Semi-automated pull request generated by
https://github.com/timgates42/meticulous/blob/master/docs/NOTE.md


TomTheBear:
Thanks for finding and fixing it.

PR #368, opened 2021-01-23
Fix perfscope when drawing y2title with multi-core

jimmysitu:
Hi, @TomTheBear 

I found there is an issue when likwid-perfscope drawing multi-core events curves
```bash
likwid-perfscope -t 1s -g L3 -c 0,4 likwid-bench -s 5s -t stream_avx -w S0:1MB:2
```
1. Only one curve align to y2
2. Core events are not align to the right core legends

I fixed this issue with the PR


TomTheBear:
Thanks for finding this and submitting a PR. I will add this to the 5.1.1 release.

PR #369, opened 2021-01-25
Fix memory measurements for AMD Zen2

TomTheBear:
Memory measurements on AMD Zen2 need to be scaled dependent on the NPS mode (`4.0/num_numadomains`). This works well for single-socket servers but for multi-socket systems, the divisor needs to be the number of NUMA domains per socket. This patch fixes this by including the number of CPU sockets in the scaling factor: `4.0/(num_numadomains/num_sockets)`.


TomTheBear:
The current state fixes the problem only partly. It works if allocation (and HW threads) are in the first NUMA domain of each socket. As soon as another NUMA domain is used, the measurements do not fit anymore.

TomTheBear:
I accidently committed the code of this PR to master. So, I close this PR.

PR #370, opened 2021-01-26
make quick install even quicker

N-Coder:


TomTheBear:
Thanks for the updated quick install.

PR #383, opened 2021-03-15
Add more structure to C-likwidAPI.c, to better match use cases

aowenson:
I have added the concept of multiple 'work items' (e.g. multiple loops), and showcase perfmon_getLastResult() to print counts for each work item. Should match better with typical use cases.

I also removed use of strtok() - as it modified 'estr', it could only be used once. Replaced with a simple array.


TomTheBear:
Looks good, thanks a lot.

PR #384, opened 2021-03-16
Add static information about the cache topology of Fujitsu A64FX

TomTheBear:


TomTheBear:
Tested on FX700 (thx @cod3monk) and FX1000 (Fugaku)

PR #385, opened 2021-03-22
Add missing build requirement for rpm specfile

blastmaster:
This commit, add the perl-Data-Dumper package as rpm build requirement.
Otherwise an rpmbuild with the provided specfile will fail.


TomTheBear:
Thanks a lot.

PR #418, opened 2021-07-22
Fix a bug due to the unrelased lock before the method returns (the issue#416)

jenny-cheung:
Fix a bug due to the unreleased lock before the method likwid_markerStopRegion returns. Insert the unlock statement before returning.


TomTheBear (approved):
Looks good.

PR #420, opened 2021-07-27
Add multi GPU capability to test/triad.cu

te42kyfo:
Starts an OpenMP thread for every GPU that is present and makes a separate benchmark for each GPU.


TomTheBear (approved):
I approve this change. There might be problems with the NvMarkerAPI since it is called by each thread now. But this is a problem of the NvMarkerAPI, not this test code. It is highly likely that we need `omp single` pragmas and barriers or such to execute the NvMarkerAPI calls only once for all GPUs in the GPUset. Or we put some logic inside the NvMarkerAPI to handle such a case.

PR #421, opened 2021-08-01
Support spaces in marker API labels

jrmadsen:
- replaces "%s" with "%Nc" where N is set to one minus
  the size of the zeroed out.
- added printout of regiontag to not valid region description msg

Hey @TomTheBear I was using the marker API and was generating some strings dynamically in timemory along the lines of: `"run/ex_cxx_overhead.cpp:314 [with timing = mode::manual]"` and was getting errors + a segfault at the end of the application like so:

```console
Line 1:run/ex_cxx_overhead.cpp:314 [with timing = mode::none]-0
 not a valid region description
Line 2:run/ex_cxx_overhead.cpp:314 [with timing = mode::manual]-0
 not a valid region description
Line 3:run/ex_cxx_overhead.cpp:314 [with timing = mode::single]-0
 not a valid region description
Line 4:run/ex_cxx_overhead.cpp:314 [with timing = mode::blank]-0
 not a valid region description
Line 5:run/ex_cxx_overhead.cpp:314 [with timing = mode::blank_pointer]-0
 not a valid region description
Line 6:run/ex_cxx_overhead.cpp:314 [with timing = mode::chained]-0
 not a valid region description
Line 7:run/ex_cxx_overhead.cpp:314 [with timing = mode::basic]-0
 not a valid region description
Line 8:run/ex_cxx_overhead.cpp:314 [with timing = mode::basic_pointer]-0
 not a valid region description
Line 9:run/ex_cxx_overhead.cpp:314 [with timing = mode::invoke]-0
 not a valid region description
Segmentation fault (core dumped)
```

Which was a bit opaque since the string looked correct. But then I added `regiontag` to the "not a valid region description" message and realized what has happening with the spaces so I patched `perfmon.c` and `nvmon.c` to support spaces and I included that diagnostic as part of this PR since it was helpful.

If I were you, I'd verify my implementation... my C string manipulation skills are fairly rusty due to my predilection for C++ strings.

### DIFF
--- a/src/nvmon.c
+++ b/src/nvmon.c
@@ -866,7 +866,7 @@ double nvmon_getMetric(int groupId, int metricId, int gpuId)
     {
         return NAN;
     }
-    
+
     char* f = ginfo->metricformulas[metricId];
     NvmonDevice_t device = &nvGroupSet->gpus[gpuId];
     if (groupId < 0 || groupId >= device->numNvEventSets)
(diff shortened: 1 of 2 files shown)


TomTheBear:
Are these two changes really enough? There is not a single change in `src/libperfctr.c` which implements the MarkerAPI.

jrmadsen:
I believe so. I didn't delve to deeply into it at the time because it fixed my issue and it was pretty clear from the message that up until these functions with the modifications were called, the string with the spaces was perfectly preserved, i.e. `"run/ex_cxx_overhead.cpp:314 [with timing = mode::invoke]"` from `Line 9:run/ex_cxx_overhead.cpp:314 [with timing = mode::invoke]-0`. Looking at `src/libperfctr.c`, it appears that is the case bc the register/start/stop region functions use `bfromcstralloc` which memcpy based on the strlen and thus the spaces are implicitly included there. 

Side note: I did not test the nvmon backend unfortunately since I was running on a system with an AMD GPU (if you weren't aware, I left NERSC and took a position in AMD Research to work on their tooling about 1.5 months ago).

PR #431, opened 2021-09-22
Improve mention of Julia interface LIKWID.jl

carstenbauer:
Now that [LIKWID.jl](https://github.com/JuliaPerf/LIKWID.jl) is really getting into shape, I suggest we update the mentions of LIKWID.jl. This PR is the first minor improvement in the README.md.

(Unfortunately, it doesn't seem to be possible to suggest changes to the WIKI - please enlighten me if there is a way - so I'll talk about those with @TomTheBear directly some time soon.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,8 +137,8 @@ Extras
 https://github.com/jacek-lewandowski/likwid-java-api
 - For Python you can find an interface to the LIKWID API here:
 https://github.com/RRZE-HPC/pylikwid or `pip install pylikwid`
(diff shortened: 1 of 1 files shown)


TomTheBear:
There is no way to suggest changes to the wiki through PRs or issues. You can clone the wiki repo and create branches but there is no way to give it back. That's unfortunately a missing feature of Github.

It depends how much you want to change. For bigger changes, I could give you repo access. For small changes it's probably easier you send me change requests as patches here or the Matrix chat.

PR #442, opened 2021-10-29
bench: avoid peakflops register dependencies

melven:
Improves results on Intel Westmere which was missing a factor of 2 before
due to latency/register dependencies in the peak flops benchmark.

### DIFF
--- a/bench/x86-64/peakflops.ptt
+++ b/bench/x86-64/peakflops.ptt
@@ -9,7 +9,6 @@ INSTR_CONST 32
 INSTR_LOOP 20
 UOPS 19
 movaps FPR1, [rip+SCALAR]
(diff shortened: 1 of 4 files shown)


melven:
Hi @TomTheBear 

This reverses some of your changes from 2 years ago.

I didn't get the "correct" 2 flops/cycle (mult+add) for scalar operations on a Westmere, this improves it (but also only results in 1.75 flops/cycle on Westmere).

But I just tried to improve it without real insight into the .ptt format.

TomTheBear:
Thanks for the PR. So, it's just reordering of the instructions and registers?

PTT format is tricky. I tried to document it here: https://github.com/RRZE-HPC/likwid/wiki/Likwid-Bench#adding-benchmarks

PR #444, opened 2021-12-01
Support for ARM Neoverse N2

TomTheBear:
- [x] Event list based on https://developer.arm.com/documentation/102099/0000/Performance-Monitors-Extension-support-/Performance-monitors-events
- [x] Counter list with 6 counters per hardware thread
- [ ] Topology detection (`/proc/cpuinfo`)
- [ ] perf_event unit name(s)


TomTheBear:
If anyone has access to such a system:
- Output of `/proc/cpuinfo`
- File listing `ls /sys/devices`

TomTheBear:
Unfortunately, cannot merge this PR as information is still missing.

breiters:
Hi Thomas, see the files attached below. Greetings from Munich.

[cpuinfo.txt](https://github.com/user-attachments/files/18801686/cpuinfo.txt)
[lscpu.txt](https://github.com/user-attachments/files/18801688/lscpu.txt)
[sys_devices.txt](https://github.com/user-attachments/files/18801689/sys_devices.txt)

Issue #481, opened 2022-08-22
[Build] Issue with ARMv8 Preprocessor Guards


OoJJBoO:
While trying to compile likwid 5.2.2 on a Raspberry Pi4 using either the GCCARMv7 or GCCARMv8 compiler options I encountered an issue regarding the checks used throughout the code: 

It looks like the code most of the time looks at `__ARM_ARCH_8A` to check if we are building using an ARMv8 compiler. But at least my version of GCC 10.4.0 built from source seems to use `__ARM_ARCH_8A__` instead. This, I think, causes the build to try to make x86 code which then causes compilation errors. The first encountered file where this is the case is `includes/cpuid.h` which results in:

```
COMPILE  GCCARMv7/frequency_uncore.o
In file included from ./src/includes/frequency.h:41,
                 from ./src/frequency_uncore.c:42:
./src/frequency_uncore.c: In function ‘freq_setUncoreFreqMin’:
./src/includes/cpuid.h:48:5: error: impossible constraint in ‘asm’
   48 |     __asm__ volatile(".ifnc %%ebx,%3 ; movl  %%ebx,%3 ; .endif  \n\t" \
      |     ^~~~~~~
./src/includes/frequency.h:49:5: note: in expansion of macro ‘CPUID’
   49 |     CPUID(eax,ebx,ecx,edx);
      |     ^~~~~
make: *** [Makefile:302: GCCARMv7/frequency_uncore.o] Error 1
```

When replacing every instance of `__ARM_ARCH_8A` inside the source code with `__ARM_ARCH_8A__` these errors do not appear. I did not thouroughly investigate if I broke anything doing that, but the resulting ARMv7 installation (at least the supported parts of it) seems to work fine. I could not test ARMv8 yet, since I can not use the default `-mabi=lp64` compiler option set in `make/include_GCCARMv8` which looks to not be supported on this system and changing it results in a linker error.

---

**Which version do you want to build?**
likwid 5.2.2 (latest stable release as of now)

**Which architecture do you want to build for?**
- [ ] x86_64
- [x] ARMv8
- [ ] POWER9
- [ ] Nvidia GPUs
- [ ] x86 (32 bit)
- [x] ARMv7
- [ ] POWER8

**To Reproduce**
- set compiler option to either GCCARMv7 or GCCARMv8
- set used compiler to GCC 10.4.0 built from source with ARMv8 target

TomTheBear:
[...]

PR #482, opened 2022-08-23
Add architecture definition strings for ARMv7 and ARMv8 compilers

OoJJBoO:
Add architecture definition strings for ARMv7 and ARMv8 to their corresponding `make/include_*.mk` files to prevent build related issues like #481.


TomTheBear:
Thanks!

Issue #483, opened 2022-08-27
[BUG] perfctr does not detect Cortex A53 correctly in ARMv7 mode


OoJJBoO:
## Issue
I noticed that perfctr does not seem to work currently when run on a Raspberry Pi 3 Model B+ even though the Cortex A53 processor (running in ARMv7 mode) should be supported as far as I understand the documentation. Running `likwid-perfctr -e` results in:
```
ERROR - [./src/perfmon.c:perfmon_init_maps:1322] Unsupported ARMv7 Processor
ERROR - [./src/perfmon.c:perfmon_check_counter_map:746] Counter and event maps not initialized.
This architecture has 0 counters.
Counter tags(name, type<, options>):



This architecture has 0 events.
Event tags (tag, id, umask, counters<, options>):
```

## Potential Fix
While looking a bit around the source, I found that in `topology.c` and `perfmon.c` the switches in lines 1137 and 1287 respectively both use `cpuid_info.model` which in my case is 0 and not `cpuid_info.part` which looks to contain the correct value defined in `ARM_CORTEX_A53`. When building after changing this, the output of `likwid-perfctr -e` changes to the following:
```
This architecture has 0 counters.
Counter tags(name, type<, options>):



This architecture has 0 events.
Event tags (tag, id, umask, counters<, options>):
```

Still, no counters were detected. Then I noticed that the path that is set for the PMC for the Cortex A53 in `includes/perfmon_a57_counters.h`, namely `/sys/bus/event_source/devices/cpu`, does not exist on my machine. After changing it to `/sys/bus/event_source/devices/armv7_cortex_a7` and rebuilding, perfctr now looks to be working:
```
This architecture has 6 counters.
Counter tags(name, type<, options>):
PMC0, Core-local general purpose counters
PMC1, Core-local general purpose counters
PMC2, Core-local general purpose counters
PMC3, Core-local general purpose counters
PMC4, Core-local general purpose counters
PMC5, Core-local general purpose counters



This architecture has 84 events.
Event tags (tag, id, umask, counters<, options>):
SW_INCR, 0x0, 0x0, PMC
L1I_CACHE_REFILL, 0x1, 0x0, PMC
L1I_TLB_REFILL, 0x2, 0x0, PMC
L1D_CACHE_REFILL, 0x3, 0x0, PMC
L1D_CACHE, 0x4, 0x0, PMC
L1D_TLB_REFILL, 0x5, 0x0, PMC
[...]

PR #484, opened 2022-08-27
Potential fix for likwid-perfctr not working on Cortex A53 when running in ARMv7 mode

OoJJBoO:
Potential fix for the issue as described in #483.

### DIFF
--- a/src/perfmon.c
+++ b/src/perfmon.c
@@ -1284,7 +1284,7 @@ perfmon_init_maps(void)
 #endif
 
         case ARMV7_FAMILY:
-            switch ( cpuid_info.model )
+            switch ( cpuid_info.part )
             {
                 case ARMV7L:
                 case ARM7L:
--- a/src/topology_hwloc.c
+++ b/src/topology_hwloc.c
@@ -284,21 +284,6 @@ hwloc_init_cpuInfo(cpu_set_t cpuSet)
         cpuid_info.stepping = atoi(info);
     snprintf(cpuid_info.architecture, 19, "x86_64");
 #endif
-/*
-#ifdef __ARM_ARCH_7A__
-    if ((info = LIKWID_HWLOC_NAME(obj_get_info_by_name)(obj, "CPUArchitecture")))
-       cpuid_info.family = atoi(info);
-    if ((info = LIKWID_HWLOC_NAME(obj_get_info_by_name)(obj, "CPURevision")))
-        cpuid_info.model = atoi(info);
-    if (cpuid_info.family == 0 || cpuid_info.model == 0)
-    {
-        uint32_t part = 0;
-        parse_cpuinfo(&count, &cpuid_info.family, &cpuid_info.model, &cpuid_info.stepping, &cpuid_info.part, &cpuid_info.vendor);
-        parse_cpuname(cpuid_info.osname);
-    }
-    snprintf(cpuid_info.architecture, 19, "armv7");
-#endif
-*/
 #if defined(__ARM_ARCH_7A__) || defined(__ARM_ARCH_8A)
     parse_cpuinfo(&count, &cpuid_info.family, &cpuid_info.model, &cpuid_info.stepping, &cpuid_info.part, &cpuid_info.vendor);
     parse_cpuname(cpuid_info.osname);
--- a/src/topology_hwloc.c
+++ b/src/topology_hwloc.c
@@ -300,7 +300,6 @@ hwloc_init_cpuInfo(cpu_set_t cpuSet)
 #endif
 */
 #if defined(__ARM_ARCH_7A__) || defined(__ARM_ARCH_8A)
-    /* uint32_t part = 0; */
     parse_cpuinfo(&count, &cpuid_info.family, &cpuid_info.model, &cpuid_info.stepping, &cpuid_info.part, &cpuid_info.vendor);
     parse_cpuname(cpuid_info.osname);
 #ifdef __ARM_ARCH_7A__
--- a/src/topology_hwloc.c
+++ b/src/topology_hwloc.c
@@ -284,6 +284,7 @@ hwloc_init_cpuInfo(cpu_set_t cpuSet)
         cpuid_info.stepping = atoi(info);
     snprintf(cpuid_info.architecture, 19, "x86_64");
 #endif
+/*
 #ifdef __ARM_ARCH_7A__
     if ((info = LIKWID_HWLOC_NAME(obj_get_info_by_name)(obj, "CPUArchitecture")))
        cpuid_info.family = atoi(info);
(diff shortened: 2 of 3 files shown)


TomTheBear:
Thanks for the PR. I have never put my hands on a RasPi3, only RasPi4.

I think the fixes are too far reaching. There are ARMv7 systems working with the current setup, so changing the architecture detection to only detect the RasPi3 is probably disables support for other systems. I would recommend to add a new `raspi3_translate_types` with your config. Also create a separate `switch case` part for RasPi3. If you get problems due to further architecture checks in the code, change the `cpuid_info.model` to `cpuid_info.part` in your `case` statement. There are two `switch case` statements in `src/perfmon.c`, one to select "lists" (event list, counter list, ...) and one to select the "functions". Remember to change it in both locations, there you can also adjust the `translate_types` list to point to the new RasPi3 `perf` types.

OoJJBoO:
> I think the fixes are too far reaching. There are ARMv7 systems working with the current setup, so changing the architecture detection to only detect the RasPi3 is probably disables support for other systems. 

Yeah, I already figured that this would probably break things for other systems.

> I would recommend to add a new `raspi3_translate_types` with your config. Also create a separate `switch case` part for RasPi3. If you get problems due to further architecture checks in the code, change the `cpuid_info.model` to `cpuid_info.part` in your `case` statement. There are two `switch case` statements in `src/perfmon.c`, one to select "lists" (event list, counter list, ...) and one to select the "functions". Remember to change it in both locations, there you can also adjust the `translate_types` list to point to the new RasPi3 `perf` types.

I will look into that, thank you!

OoJJBoO:
To understand how that is done in the source code, I now have gone through the code that sets the values stored in `cpuid_info` for ARM processors. The thing is, while doing that, I came to the conclusion that it looks like for every ARMv7 or ARMv8 processor `cpuid_info.model` will end up as the value listed under `CPU variant` in `/proc/cpuinfo`. This does, at least as far as I understand it, not represent the actual cpu model and therefore should probably not be used in the switches I edited earlier.

This is caused by `hwloc_init_cpuInfo` in `topology_hwloc.c` overwriting the value that was set in lines 287 - 301 which read
```c
#ifdef __ARM_ARCH_7A__
    if ((info = LIKWID_HWLOC_NAME(obj_get_info_by_name)(obj, "CPUArchitecture")))
       cpuid_info.family = atoi(info);
    if ((info = LIKWID_HWLOC_NAME(obj_get_info_by_name)(obj, "CPURevision")))
        cpuid_info.model = atoi(info); // <-- value was first set to 3 in my case here
    if (cpuid_info.family == 0 || cpuid_info.model == 0)
    {
        uint32_t part = 0;
        parse_cpuinfo(&count, &cpuid_info.family, &cpuid_info.model, &cpuid_info.stepping, &cpuid_info.part, &cpuid_info.vendor);
        parse_cpuname(cpuid_info.osname);
    }
    snprintf(cpuid_info.architecture, 19, "armv7");
#endif
```

in the next code block that looks like
```c
#if defined(__ARM_ARCH_7A__) || defined(__ARM_ARCH_8A) // <-- should the first define be there?
    uint32_t part = 0;
    parse_cpuinfo(&count, &cpuid_info.family, &cpuid_info.model, &cpuid_info.stepping, &cpuid_info.part, &cpuid_info.vendor); // <-- value is updated to 0 here regardless of previous result
    parse_cpuname(cpuid_info.osname);
    snprintf(cpuid_info.architecture, 19, "armv8");
#endif
```

Now, `parse_cpuinfo` sets `cpuid_info.model` to the cpu variant, which in my case is 0. When looking at what I found online, this probably also affects other ARM processors, since their variants reported in `/proc/cpuinfo` will not match the values expected by the switch cases where `cpuid_info.model` is used for the `ARMV7_FAMILY` case.

For the `ARMV8_FAMILY` case this will not be an issue, since the switches already look at `cpuid_info.part` and not the model there.

So, to summarize, I think that possibly either:
1. There is a typo in line 300 of `topology_hwloc.c` and `#if defined(__ARM_ARCH_7A__) || defined(__ARM_ARCH_8A)` could maybe be raplaced with `#ifdef __ARM_ARCH_8A`. Using this small change, my CPU type gets listed as "ARM 7l architecture" by `likwid-topology` instead of "nil".
2. The switch cases for the `ARMV7_FAMILY` do not work as intendet at the moment like I suspected previously and could maybe be updated to to use `cpuid_info.part` instead of the `cpuid_info.model` or also take `cpuid_info.vendor` into account like is already done in the `ARMV8_FAMILY` case. This is the variant I previously suggested (without the changed `translate_types`).

OoJJBoO:
>     1. ... Using this small change, my CPU type gets listed as "ARM 7l architecture" by `likwid-topology` instead of "nil". ...
Sadly I have to take that back, since I accidentally tested this on my Pi4 instead of the Pi3. On the Pi3, the cpu revision is 4 which of course is not the correct value that is expected by the switches.

But this of course does not change the fact that the code I mentioned in my previous post looks a bit odd since it basically overwrites most if not all that was done in the previous code block and I still think the `ARMV7_FAMILY` cases might not quite work as intended and probably should be updated.

OoJJBoO:
I can at least confirm that the current version of my proposed change `likwid-topology` now reports the correct cpu type on both the Pi3 (Cortex A53) and Pi4 (Cortex A72). Since I do not have access to any other ARM-based machines, these are all I can test at the moment.

src/perfmon.c
```diff
@@ -1284,7 +1284,7 @@ perfmon_init_maps(void)
 #endif
 
         case ARMV7_FAMILY:
-            switch ( cpuid_info.model )
+            switch ( cpuid_info.part )
```
OoJJBoO, line 1287: The other switch does not include a case for the ARMv7 family, so this is all there is to do in this file I think.

TomTheBear:
Thanks for all this digging. The architecture detection is a mess and grown historically (LIKWID still supports Intel Pentium I and AMD K8).

The `parse_cpuinfo` shouldn't be called twice, so removing the ARMv7 define there is fine. ARMv7 is probably not part of the "function selector `switch case`" because LIKWID always uses perf_event for ARMv7 and ARMv8, so no selection required. If switching to `part` instead of `model` is enough to make the detection work, we should do that.

src/perfmon.c
TomTheBear: That's fine, the functions are fixed to the `perf_event` functions for ARM.

OoJJBoO:
>
> The `parse_cpuinfo` shouldn't be called twice, so removing the ARMv7 define there is fine.
>
The first code block does only set `cpuid_info.part` in some cases though, so removing the define will probably break things again. On the other hand, I think it should be safe to remove the first code block following the `#ifdef __ARM_ARCH_7A__` altogether since everything done there is overwritten in the next block regardless as things are now.

I updated my fork to include this change. Also, I moved the assignment of `cpuid_info.architecture` to their own defines for both ARM architectures so it actually should represent the correct one (currently it is always set to armv8). If for some reason both the `__ARM_ARCH_7A__` and `__ARM_ARCH_8A` defines are present, it will be set to armv8.

src/topology_hwloc.c
```diff
@@ -284,6 +284,7 @@ hwloc_init_cpuInfo(cpu_set_t cpuSet)
         cpuid_info.stepping = atoi(info);
     snprintf(cpuid_info.architecture, 19, "x86_64");
 #endif
+/*
```
OoJJBoO, line 287: This block effectively does nothing and should be safe to remove.

src/topology_hwloc.c
```diff
@@ -297,12 +298,18 @@ hwloc_init_cpuInfo(cpu_set_t cpuSet)
     }
     snprintf(cpuid_info.architecture, 19, "armv7");
 #endif
+*/
 #if defined(__ARM_ARCH_7A__) || defined(__ARM_ARCH_8A)
-    uint32_t part = 0;
+    /* uint32_t part = 0; */
```
OoJJBoO, line 303: This value is never used.

TomTheBear: Please remove the line, not comment it out.

OoJJBoO: Okay

TomTheBear:
Not sure whether it's your changes, the placement of the comments or Github's UI but it is hard to see the changes you have done. But I'll look into your changes.

OoJJBoO:
I just removed the code I previously commented out. Sorry if that made things a bit unclear, I don't know why I didn't do that from the start.

TomTheBear:
Looks good. Thanks.

PR #495, opened 2022-11-09
Add support for Apple M1 Studio

TomTheBear:
This PR adds the Apple M1 chips to LIKWID

- topology detection
  - thread & NUMA topology work
  - no cache information provided by the system (common on ARM). Static definition causes problems due to the thread topology
- performance monitoring
  - Only `perf_event` access mode (default with `COMPILER=GCCARMv8`)
  - 10 `PMC` counters with only a few named events
  - The selection whether the `icestorm` or `firestorm` PMU is used, is selected under the hood (might change depending on the feature set of the PMUs. At the moment, both PMU provide the same counters and events)
  - MarkerAPI works
- micro-benchmarking
  - Common `armv8` benchmarks work as long as no SVE is used
  - No kernels for M1 specific instructions yet


TomTheBear:
It seems the recent kernels used on my test system do not export `/sys/devices/system/cpu*/topology/die_id` anymore but use `/sys/devices/system/cpu*/topology/cluster_id` instead.

PR #499, opened 2022-11-29
Fix segfault when invalid affinity domain is specified

christgau:
There is an issue in the workgroup parsing code that causes a segfault (also with latest master commit):

```
$ ./likwid-bench -t store -w S0:1GB  # works
Allocate: Process running [...] 
$ ./likwid-bench -t store -w X0:1GB # does not work 
Segmentation fault (core dumped)
```
The current code checks for two elements to be presents in the token list, but accesses the non-existing third element which is fixed by this PR. The outcome then is:

``` 
$ ./likwid-bench -t store -w X0:1GB
Unknown affinity domain X0
```

### DIFF
--- a/bench/src/strUtil.c
+++ b/bench/src/strUtil.c
@@ -154,7 +154,7 @@ parse_workgroup(Workgroup* group, const_bstring str, DataType type)
         }
         else
         {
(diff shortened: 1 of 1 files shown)


TomTheBear:
Thanks for reporting and fixing.

PR #519, opened 2023-03-08
Use _exit() instead of exit() when checking for RDPMC support

JamesWrigley:
Checking for RDPMC support is done in a `fork()`'ed child process, and when the parent is a Julia process there can be exit handlers registered that will crash in the child process. We can get around that by using `_exit()` instead, which doesn't run exit handlers.

Fixes https://github.com/JuliaPerf/LIKWID.jl/issues/51 for me. AFAIK it's safe to use `_exit()` here, according to the [docs](https://man7.org/linux/man-pages/man2/_exit.2.html) the only things to be wary of are stdio streams not being flushed (but we don't print anything in the child) and open file descriptors being closed (I can't see any file descriptors used in the child). I can't see any difference in signal handling either.

CC @carstenbauer

### DIFF
--- a/src/access_x86_rdpmc.c
+++ b/src/access_x86_rdpmc.c
@@ -89,7 +89,7 @@ __rdpmc(int cpu_id, int counter, uint64_t* value)
 void
 segfault_sigaction_rdpmc(int signal, siginfo_t *si, void *arg)
 {
-    exit(1);
+    _exit(1);
 }
 
 static int
@@ -106,6 +106,10 @@ test_rdpmc(int cpu_id, uint64_t value, int flag)
     }
     if (!pid)
     {
+        // Note: when exiting the child process we use _exit() instead of exit()
+        // to avoid calling exit handlers registered by the parent process
+        // (e.g. libuv functions from Julia), which can cause crashes.
+
         uint64_t tmp;
         struct sigaction sa;
         memset(&sa, 0, sizeof(struct sigaction));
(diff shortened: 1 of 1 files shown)


TomTheBear:
Thanks a lot for the fix.

The man page says it returns `status & 0xFF` (my system: `status & 0377`) to the parent process. That's the only questionable part for me. If the `WIFEXITED` and `WEXITSTATUS` macros still works, I'm fine with this fix.

JamesWrigley:
I think that's doing the same thing as `exit()`? According to the [docs](https://man7.org/linux/man-pages/man3/exit.3.html):
> The exit() function causes normal process termination and the
       least significant byte of status (i.e., status & 0xFF) is
       returned to the parent (see [wait(2)](https://man7.org/linux/man-pages/man2/wait.2.html)).

TomTheBear:
That's also my assumption. Ready for merge.

jrmadsen:
If I understand the context correctly, I think you might want to use [quick_exit](https://en.cppreference.com/w/c/program/quick_exit) since that causes normal termination but doesn't clean up any resources. Theoretically, if you are using LIKWID on a C++ code, the code could use the destructors of static objects instead of atexit handlers to clean up resources. I've seen the destructor of `std::thread` within child forked processes cause problems bc the thread object is copied but the thread itself is not. That's *probably* especially true for the new `std::jthread` which will try to join the thread during destruction if it is still running.

JamesWrigley:
I'm not an expert, but makes sense to me and it also fixes my issue :+1: Opened a PR at #520.

PR #520, opened 2023-03-08
Use quick_exit() instead of _exit()

JamesWrigley:
Inspired by https://github.com/RRZE-HPC/likwid/pull/519#issuecomment-1460605455.


TomTheBear:
Thanks @jrmadsen .

PR #543, opened 2023-07-26
Add vectorization ratio to MEM_*P group where possible

### LINKED ISSUE #540: Add vectorization ratio to MEM_DP group where possible
Vectorization ratio is only printed with FLOPS_DP (and FLOPS_SP), not with the MEM_* groups on Intel CPUs. I'd like to see it in all groups where it can be calculated.

LIKWID 5.2.2

TomTheBear:
Fixes #540


TomTheBear:
See #553

PR #554, opened 2023-09-20
Add percentage unit to vectorization ratio

### LINKED ISSUE #553: Add the % sign for metrics which represent a percentage
I noticed [here](https://github.com/RRZE-HPC/likwid/blob/master/groups/CLX/FLOPS_DP.txt#L22) that vectorization ratio is computed as a percentage but it is not displayed as such. To avoid confusion, I would suggest to display it with a little `[%]` on the side.

TomTheBear:
Fixes #553


TomTheBear:
```
$ for F in $(find . -name "*.txt"); do sed -i -e "s/Vectorization ratio/Vectorization ratio [%]/g" $F; done
```

PR #570, opened 2023-11-01
Adopt shebang lines because of build errors

skriesch:
The rpm runtime does not like the syntax of some shebang lines in the code. We are receiving error messages:

```
[   63s] likwid.x86_64: E: env-script-interpreter (Badness: 9) /usr/bin/feedGnuplot /usr/bin/env perl
[   63s] likwid.x86_64: E: env-script-interpreter (Badness: 9) /usr/share/likwid/filter/json /usr/bin/env python3
[   63s] likwid.x86_64: E: env-script-interpreter (Badness: 9) /usr/share/likwid/filter/template /usr/bin/env perl
[   63s] likwid.x86_64: E: env-script-interpreter (Badness: 9) /usr/share/likwid/filter/xml /usr/bin/env perl
[   63s] This script uses 'env' as an interpreter. For the rpm runtime dependency
[   63s] detection to work, the shebang #!/usr/bin/env <interpreter>  needs to be
[   63s] patched into #!/usr/bin/<interpreter>  otherwise the package dependency
[   63s] generator merely adds a dependency on /usr/bin/env rather than the actual
[   63s] interpreter /usr/bin/<interpreter>.  Alternatively, if the file should not be
[   63s] executed, then ensure that it is not marked as executable or don't install it
[   63s] in a path that is reserved for executables.
```

Therefore, I have adopted the referenced files as the rpm runtime (in the openSUSE Build Service) is suggesting for more successful builds in the future.

### DIFF
--- a/filters/json
+++ b/filters/json
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/python3
 
 import json
 import sys
--- a/filters/template
+++ b/filters/template
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/bin/perl
 
 use strict;
 use warnings;
--- a/filters/xml
+++ b/filters/xml
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/bin/perl
 
 use strict;
 use warnings;
--- a/perl/feedGnuplot
+++ b/perl/feedGnuplot
@@ -1,4 +1,4 @@
-#!/usr/bin/env perl
+#!/usr/bin/perl
 
 package feedgnuplot; # for the metacpan indexer
 


TomTheBear:
Thanks for PR. I'm not sure whether this should be "fixed" globally or applied separately as patch when creating RPMs.

On HPC systems with `modules` or `spack env`, there are often no fixed paths that can be used in the shebang. The `/usr/bin/env <interpreter>` approach takes the `<interpreter>` currently available in the environment. I admit, `perl` and `python3` should be there by default.

Can you please put it in the `likwid.spec` file (with `sed` or something) and use the RPM macros like `%{__python}` instead of hardcoded paths?

TomTheBear:
Thanks for fixing it in the spec file. There should be a macro for `perl` as well. It should be supported when `perl` is a build requirement which it should because LIKWID uses Perl scripts in the build phase.

You can pass arguments to the interpreter through env:
```
#!/usr/bin/env -S <interpreter> <arg1> <arg2>
```
For more info, see [here](https://www.gnu.org/software/coreutils/manual/html_node/env-invocation.html#g_t_002dS_002f_002d_002dsplit_002dstring-usage-in-scripts)

Issue #583, opened 2023-11-13
[BUG] Perfmon marker file API functions not working


OoJJBoO:
## Describe the bug

As of commit 1a50c17, the perfmon functions provided for working with marker files (`perfmon_readMarkerFile`, `perfmon_getNumberOfRegions`, `perfmon_getEventsOfRegion`, and so on) are effectively broken.

This is because they, as far as I understand, need `likwid_markerClose` to be called for the marker file to actually be written first, which, since the mentioned commit, also finalizes the performance monitoring facilities. This, in turn, causes `perfmon_readMarkerFile` to always fail with the error
```
ERROR - [...] Perfmon module not properly initialized
```
when trying to read the created marker file afterward because perfmon was already finalized and the needed data does not exist anymore.

## To Reproduce

To reproduce, simply run the provided `C-internalMarkerAPI` example or try calling `perfmon_readMarkerFile` after `likwid_markerClose` has written a marker file.

## Possible Fix / Workaround

Since it looks to be complicated to remove the `perfmon_finalize` call from `likwid_markerClose`, I think another possible way to address this issue would be to add a new function that only writes the marker file without erasing the performance monitoring data.  That way, users would be able to write the marker file manually and subsequently read results using the existing API functions.

I already implemented this approach in [my local fork](https://github.com/OoJJBoO/likwid/tree/save_marker_file_func) and it seems to be working so far after some quick experiments. The `likwid_writeMarkerFile` function I added is a modified copy of the relevant code in `likwid_markerClose` with perfmon calls replacing the `LikwidResults` object created by `hashTable_finalize`.

Output of  `C-internalMarkerAPI` using the current repository state (commit b611b8e):
```
gcc -g -fopenmp -DLIKWID_PERFMON -I/usr/local/include -L/usr/local/lib C-internalMarkerAPI.c -o C-internalMarkerAPI -llikwid -lm
LD_LIBRARY_PATH=/usr/local/lib ./C-internalMarkerAPI
calc_flops thread 2 finished measuring group FLOPS_DP.
Got 6 events, runtime 1.339779 s, and call count 1
calc_flops thread 0 finished measuring group FLOPS_DP.
Got 6 events, runtime 1.340276 s, and call count 1
calc_flops thread 1 finished measuring group FLOPS_DP.
Got 6 events, runtime 1.340675 s, and call count 1
detailed event results: 
                       INSTR_RETIRED_ANY:              4500023470.000000
[...]

PR #584, opened 2023-11-13
Add function for manually writing marker file

OoJJBoO:
Add `likwid_writeMarkerFile` function to address the issue described in #583.


TomTheBear:
Looks good, thanks for your work on the comments.

I found a small error in the F90 interface. The function in `likwid_f90_interface.c` has to have an underscore in the end: `likwid_writemarkerfile_`.

OoJJBoO:
Updated. But looking at the other functions in the same file, I guess I should probably also use the way of copying the string first before passing it to the `likwid_writeMarkerFile` function, right? Didn't notice that earlier for some reason.

TomTheBear:
Yes, you should copy the string before providing it to the C functions.

OoJJBoO:
Okay, I changed it. I also added the missing register macro in the Lua bindings.

TomTheBear:
Sorry that I didn't check more thoroughly. I'm on a business trip (SC23) and read your code in between meetings. Next week I have more time to go through your changes.

OoJJBoO:
No worries, take your time :)

TomTheBear:
I looked at your code again and I realized that the new functions/macros do not follow the current naming scheme:
- `likwid_writeMarkerFile` -> `likwid_markerWriteFile`
- `LIKWID_WRITE_MARKER_FILE` -> `LIKWID_MARKER_WRITE` or `LIKWID_MARKER_WRITE_FILE`
- `likwid_writemarkerfile_` -> `likwid_markerwritefile_`
- `lua_likwid_writeMarkerFile` -> `lua_likwid_markerWriteFile`

OoJJBoO:
Ah, I see, you're right. I'll update them soon

TomTheBear:
Thanks, that should be all functions for the CPU MarkerAPI.

I should probably add similar functions for the Nvidia MarkerAPI and the ROCm MarkerAPI.

PR #585, opened 2023-12-07
Add support for Nvidia Grace

TomTheBear:
- Added units:
  - [x] hwthread-local PMU `armv8_pmuv3_0`
  - [x] `nvidia_scf_pmu_0`
  - [x] `nvidia_pcie_pmu_0`
  - [x] `nvidia_nvlink_c2c0_pmu_0`
  - [x] `nvidia_cnvlink_pmu_0` (only Grace-Hopper)
  - [x] `nvidia_nvlink_c2c1_pmu_0` (only Grace-Hopper)
  
This PR also contains some fixes for the `COMPILER` setting `GCCARM`.

Documentation: https://docs.nvidia.com/grace-performance-tuning-guide.pdf
Uncore events based on `perf` data in `/sys/devices/nvidia*/events`


TomTheBear:
https://developer.arm.com/documentation/102375/latest/

TomTheBear:
We identified a problem on GraceGrace systems with the current PR. The memory controller devices `SCF*` cover only the first socket. The second socket is not yet supported.

Thanks @JanLJL for finding this.

TomTheBear:
The last commit fixes the Uncore device handling with `perf_event` mode on GraceGrace systems. Commonly one Uncore unit covers all sockets but in case of Nvidia GraceGrace, there are separate devices per socket.

Moreover, the GraceGrace test system reports CPU socket IDs `36` and `2364` which confused the lock setup.

PR #594, opened 2024-01-03
Add Performance Groups for Caches to Intel SapphireRapids

chriswasser:
Hi @TomTheBear,

This PR adds the two performance groups `L2CACHE` and `L3CACHE` to the Intel SapphireRapids architecture. These groups also exist for previous architectures and were adapted from the existing group files. The following adoptions were made:

- In SapphireRapids, the `L2_TRANS_ALL_REQUESTS` event of the `L2CACHE` group no longer exists and was replaced by the `L2_REQUEST_ALL` event. For consistency within the group file `L2_RQSTS_MISS` was replaced by its alias `L2_REQUEST_MISS`.
- As [this comment](https://github.com/RRZE-HPC/likwid/blob/master/groups/ICL/L3CACHE.txt#L35-L36) ("With Intel Icelake, the retired UOPs cannot be measured anymore but instead, [...]") indicates, SapphireRapids no longer supports the counting of all retired UOPs. Therefore, the `L3CACHE` group uses the event `INSTR_RETIRED_ANY` as a basis for the rates. This matches the calculation in the `L2CACHE` group although it deviates from [the calculation for Icelake](https://github.com/RRZE-HPC/likwid/blob/master/groups/ICL/L3CACHE.txt) which used the `UOPS_RETIRED_SLOTS` event.

The new groups were tested with the `stream` test from the LIKWID benchmark suite. As a sanity check the counter values were compared with enabled and disabled prefetchers as well as on SkylakeX and SapphireRapids. The reported values seemed sensible while the two architectures exhibited the qualitatively same behavior.

Hope this is helpful and can be integrated into LIKWID. Thanks for the consideration 🤓

Greetings

Christian

### DIFF
--- a/groups/SPR/L2CACHE.txt
+++ b/groups/SPR/L2CACHE.txt
@@ -1,0 +1,35 @@
+SHORT L2 cache miss rate/ratio
+
+EVENTSET
+FIXC0 INSTR_RETIRED_ANY
+FIXC1 CPU_CLK_UNHALTED_CORE
+FIXC2 CPU_CLK_UNHALTED_REF
+FIXC3 TOPDOWN_SLOTS
+PMC0  L2_REQUEST_ALL
+PMC1  L2_REQUEST_MISS
+
+METRICS
+Runtime (RDTSC) [s] time
+Runtime unhalted [s] FIXC1*inverseClock
+Clock [MHz]  1.E-06*(FIXC1/FIXC2)/inverseClock
+CPI  FIXC1/FIXC0
+L2 request rate PMC0/FIXC0
+L2 miss rate PMC1/FIXC0
+L2 miss ratio PMC1/PMC0
+
+LONG
+Formulas:
+L2 request rate = L2_REQUEST_ALL/INSTR_RETIRED_ANY
+L2 miss rate = L2_REQUEST_MISS/INSTR_RETIRED_ANY
+L2 miss ratio = L2_REQUEST_MISS/L2_REQUEST_ALL
+-
+This group measures the locality of your data accesses with regard to the
+L2 cache. L2 request rate tells you how data intensive your code is
+or how many data accesses you have on average per instruction.
+The L2 miss rate gives a measure how often it was necessary to get
+cache lines from memory. And finally L2 miss ratio tells you how many of your
+memory references required a cache line to be loaded from a higher level.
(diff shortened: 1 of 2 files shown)


TomTheBear:
Why didn't you take the events used by Intel in it's metrics? Example (l2_mpi == L2 miss rate): https://github.com/intel/perfmon/blob/main/SPR/metrics/sapphirerapids_metrics.json#L222-L244

Please add a comment that you use different events to the `LONG` section. In the L3CACHE group with `INSTR_RETIRED_ANY` but uop events for the hits/misses, you can get values > 100%.

chriswasser:
### Ad 1

`L2_LINES_IN.ALL` seems to measure cache lines not the number of requests. In the `L3` performance groups it is used to measure the data volume instead of the number of L2 misses. Since this has remained the same for a dozen or so intel architectures, I wanted to remain consistent with previous groups. A quick test exhibits a higher value for `L2_LINES_IN_ALL` than for `L2_REQUEST_ALL` which does not seem sensible to me:

```txt
|     L2_REQUEST_ALL    |   PMC0  |   312527000 |
|    L2_REQUEST_MISS    |   PMC1  |   312525300 |
|    L2_LINES_IN_ALL    |   PMC2  |   312529800 |
```

Therefore, I would advocate to keep the succinctly named `L2_REQUEST_MISS` event.

### Ad 2

I read "Counts retired load instructions with at least one uop that hit in the L3 cache."  for the `MEM_LOAD_RETIRED.L3_HIT` event (see: [here](https://github.com/intel/perfmon/blob/main/SPR/events/sapphirerapids_core.json#L7472-L7497) similarly for the `MEM_LOAD_RETIRED.L3_MISS` event) and thought it would count only once for each instruction and not multiple times if the instructions decodes to multiple micro-operations. Or is this wrongly documented / easily misunderstood?

Of course, I could still add a comment about the changed basis for the rates as done in the Icelake groups if you would want this change documented.

TomTheBear:
I would say it is not well documented. The same event code refers to uops until Broadwell. Since Skylake, it uses the instruction count according to the docs. Published metrics like [this one](https://github.com/intel/perfmon/blob/main/SPR/metrics/sapphirerapids_metrics.json#L245-L267) combine the `MEM_LOAD_*` event with instructions, so it is likely that the events really count instructions now.

Issue #599, opened 2024-01-24
[BUG] perfctr crashes on a64fx


jdomke:
**Describe the bug**
likwid-perfctr throws different Aborted (core dumped) errors depending on runtime of the sleep command

```
 $ likwid-perfctr -C 0 -g L2 sleep 1
--------------------------------------------------------------------------------
CPU name:
CPU type:       Fujitsu A64FX
CPU clock:      0.00 GHz
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
malloc(): unaligned tcache chunk detected
[1]+  Aborted                 (core dumped) likwid-perfctr -C 0 -g L2 sleep 1
Aborted (core dumped)
```

```
$ likwid-perfctr -C 0 -g L2 sleep 2
------------------------------------------------------------------------------
--
CPU name:
CPU type:       Fujitsu A64FX
CPU clock:      0.00 GHz
--------------------------------------------------------------------------------
--------------------------------------------------------------------------------
Group 1: L2
+------------------+---------+------------+
<<snip>>
|    L1<->L2 data volume [GBytes]    |     0.0020 |
+------------------------------------+------------+

double free or corruption (out)
Aborted (core dumped)
```

**To Reproduce**
* LIKWID command and/or API usage
** see above
* LIKWID version and download source (Github, FTP, package manger, ...)
** v5.3.0 tag compiled with GCCARMv8 and ACCESSMODE=direct
* Operating system
** RHEL 8.8 (Ootpa)
* Does your application use libraries like MPI, OpenMP or Pthreads?
* In case of Nvidia GPUs, which CUDA version?
* Are you using the MarkerAPI (CPU code instrumentation) or the NvMarkerAPI (Nvidia GPU code instrumentation)?

**To Reproduce with a LIKWID command**
Please supply the output of the command with `-V 3` added to the command:
* likwid-perfctr
```
$ likwid-perfctr -V 3 -C 0 -g L2 sleep 1
DEBUG - [hwloc_init_cpuInfo:367] HWLOC CpuInfo Family 8 Model 1 Stepping 0 Vendor 0x46 Part 0x1 isIntel 0 numHWThreads 24 activeHWThreads 24
DEBUG - [proc_init_nodeTopology:713] PROC Thread Pool PU 0 Thread 0 Core 0 Die 0 Socket 0 inCpuSet 1
[...]

PR #601, opened 2024-02-01
fixing two bugs which result in crashes on a 24-core a64fx (related to issue #599)

jdomke:
bug 1: numCachesPerSocket is 0 and numberOfDomains is therefore calculated incorrectly resulting in writes to not allocated mem

bug 2: 24-core version of a64fx is essentiually a 48 core with some cores being disabled, but the core IDs are not numbered consecutively by the kernel resulting in segfaults in likwid

### DIFF
--- a/src/topology_proc.c
+++ b/src/topology_proc.c
@@ -602,6 +602,7 @@ proc_init_nodeTopology(cpu_set_t cpuSet)
     int (*ownatoi)(const char*);
     ownatoi = &atoi;
     int last_socket = -1;
+    int last_coreid = -1;
     int num_sockets = 0;
     int num_cores_per_socket = 0;
     int num_threads_per_core = 0;
@@ -631,6 +632,7 @@ proc_init_nodeTopology(cpu_set_t cpuSet)
             {
                 num_sockets++;
                 last_socket = packageId;
+                last_coreid = -1;
             }
             fclose(fp);
         }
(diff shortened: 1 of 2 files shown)


src/topology_proc.c
```diff
@@ -639,7 +641,7 @@ proc_init_nodeTopology(cpu_set_t cpuSet)
         if (NULL != (fp = fopen (bdata(file), "r")))
         {
             bstring src = bread ((bNread) fread, fp);
-            hwThreadPool[i].coreId = ownatoi(bdata(src));
+            hwThreadPool[i].coreId = (++last_coreid);
```
TomTheBear, line 644: Why using a simple counter and not the content of the `core_id` file from sysfs? LIKWID historically used a lot of counters for HWthreads, NUMA nodes and other topological entities but this has been proven to be error-prone. We changed a lot of code in the past to use counters less frequently and rely on OS info.

jdomke: because core_id on that system is not 0, 1, 2, ..., 23; but instead has holes like 0, 2, 3, 4, 7, ...., 48 depending on which cores have been disabled in the CPU's "firmware". the issue is that some of likwid code allocates arrays based on the number of cores, and uses core_id as index, meaning you get out-of-bounds access which i tracked down with valgrind to see why likwid was segfaulting on that system. usually the OS takes care of the remapping of core_id into a consecutive number range, but for some reason what kernel fix is not available for this system/cpu. i admit that it is a very rare bug...

jdomke: sorry for my clumbsy explanation, but just to give you an idea, here is how it actually looks on the system:
```
⇒  cat /sys/devices/system/cpu/cpu*/topology/core_id | sort -u
0
1
10
11
5
6
7
8
```
instead of counting 0,...,5 for each CMG, you get 0,...,11 with missing numbers

PR #608, opened 2024-02-06
allow user-supplied cfg and topo files via env var

jdomke:
and in the process i cleaned up init_configuration a bit


jdomke:
thanks. sorry, i have not tried the accessdaemon (forgot why, think it wasn't properly building on that system). will check in the next few days, but it might take some time due to travel and other commitments

PR #627, opened 2024-07-10
Add support for ARM Cortex A76 reusing Neon N1 counters and events

OoJJBoO:
Since I am currently trying to use Likwid on a Raspberry Pi 5 which features an ARM Cortex A76 that is not supported yet, I added the topology definitions to `topology.h` and `topology.c` and also the counter + event definitions. Since they look to be identical, I reused the Neon 1 counter and event setup like was done for the Cortex A72 which uses the Cortex A57 definitions.


TomTheBear:
Thanks. Looks good.

PR #640, opened 2024-11-04
Add support AMD Zen4c (Bergamo)

lub-dub:
The ZEN4_EPYC_BERGAMO family was missing in the if statements, this adds them so the power monitoring also works on this platform. According tot he programming guide the registers etc are all the same.


TomTheBear:
Thanks for the PR.

Can you provide a link to the PPR for Bergamo? You added Bergamo also to the perfmon module at the other AMD Zen4 models. We have to check whether the events and counters are really similar to the other models.

lub-dub:
Sure, see https://www.amd.com/content/dam/amd/en/documents/epyc-technical-docs/programmer-references/57228.zip

TomTheBear:
My hope was that there are no changes between AMD Zen4 and Zen4c but unfortunately, the counter management changed slightly and there are some different events. The commit contains everything to use `likwid-perfctr`.

PR #645, opened 2024-11-11
Add Icelake L3CACHE group to IcelakeX

chriswasser:
Hi everyone,

just as the title reads, this PR adds the existing L3CACHE group from ICL to ICX as well. The events work the same on both architectures (see [ICL here](https://github.com/intel/perfmon/blob/dbd3a0054701814c0f833df641410d0647b05184/ICL/events/icelake_core.json) and [ICX here](https://github.com/intel/perfmon/blob/dbd3a0054701814c0f833df641410d0647b05184/ICX/events/icelakex_core.json)); just seems like a minor oversight when these groups were initially added.

Thanks for integrating this change. Greetings

Christian

### DIFF
--- a/groups/ICX/L3CACHE.txt
+++ b/groups/ICX/L3CACHE.txt
@@ -1,0 +1,37 @@
+SHORT L3 cache miss rate/ratio
+
+EVENTSET
+FIXC0 INSTR_RETIRED_ANY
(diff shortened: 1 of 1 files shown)


TomTheBear:
Thx!

PR #651, opened 2024-11-26
Add stream_sp_mem kernel

fairydreaming:
This PR adds stream_sp_mem kernel. The kernel is based on stream_sp kernel, the only difference is that movss store instructions were replaced by a combination of:
- unpcklps instructions that unpack scalar values from FPR1-FPR4 into FPR1 and values from FPR5-FPR8 into FPR5
- movntps instructions that store values from FPR1 and FPR5 into memory

Please verify the number of UOPS, as I'm not 100% sure about this.

### DIFF
--- a/bench/x86-64/stream_sp_mem.ptt
+++ b/bench/x86-64/stream_sp_mem.ptt
@@ -1,0 +1,44 @@
+STREAMS 3
+TYPE SINGLE
+FLOPS 2
+BYTES 12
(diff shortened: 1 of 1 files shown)


TomTheBear:
`UOPS` should be 32+8+2 = 42

fairydreaming:
> `UOPS` should be 32+8+2 = 42

Fixed.

PR #653, opened 2024-12-03
Fix LIKWID_VERSION and LIKWID_COMMIT macros in likwid.h not being set correctly during installation

OoJJBoO:
While playing around with different versions, I noticed that the `LIKWID_VERSION` and `LIKWID_COMMIT` macros are currently not set correctly during the installation step, because there are some discrepancies between the naming used in `likwid.h` and the scheme the corresponding `sed` command in the Makefile expects.

Using the changes in this commit, the insertion should work as originally intended.


TomTheBear:
Thanks for the PR. It has to wait until the next major release as it changes the main header `likwid.h`.

OoJJBoO:
I see, thanks anyway for the quick response :)

OoJJBoO:
Alternatively, we could probably also do it the other way around and only adjust the expected names in the Makefile, right? Would that be better @TomTheBear ?

TomTheBear:
Changing the names in the Makefile is probably a better approach. As long as no change happens to `likwid.h`, it can be merged directly and can be included in 5.4.1.

OoJJBoO:
Just replaced the first commit, this time only modifying the Makefile. Looks to also work just fine this way.

PR #667, opened 2025-03-17
Added support Emerald Rapids

vxirau:
Hello!

I followed the instructions in https://github.com/RRZE-HPC/likwid/wiki/AddX86Support to add Emerald Rapids support. 

Events are the same as for Sapphire according to [perfmon intel events](https://perfmon-events.intel.com/index.html?pltfrm=emeraldrapids_server.html). 

Let me know if there is something else that should be reviewed.


TomTheBear:
Hi, looks very good. Thanks for adding it. 

One important change is missing though. You should add it to `src/power.c` (around [here](https://github.com/RRZE-HPC/likwid/blob/master/src/power.c#L145)). Otherwise the `ENERGY` group will not work.

Some minor additions:
- For sysFeatures, add the arch to the list [here](https://github.com/RRZE-HPC/likwid/blob/master/src/sysFeatures_intel.c#L145)
- For sysFeatures rapl component, EmeraldRapids probably uses the same energy units (for conversion) as SapphireRapids, so you can add EmeraldRapids [here](https://github.com/RRZE-HPC/likwid/blob/master/src/sysFeatures_intel_rapl.c#L541) and [here](https://github.com/RRZE-HPC/likwid/blob/master/src/sysFeatures_intel_rapl.c#L672)
- For support for the fixed `TOPDOWN_SLOTS` event, you should add it to [the condition](https://github.com/RRZE-HPC/likwid/blob/master/src/perfgroup.c#L778)


Any comments on AddX86Support? Something unclear? I'm happy to update the page.

vxirau:
Hi @TomTheBear !

Thank you for your response. I just made the changes you mentioned.

The AddX86Support guide was very clear and helpful! The only thing I didn't get from the guide was the sysFeatures being something mandatory to change! But of course it was easy to add.

I've also updated the README to reflect the change. Should I update the changelog too? As mentioned in the wiki?

Thanks!

TomTheBear:
SysFeatures is a quite recent development. It is still experimental and has to be explicitly enabled at build time.  I added it to the page but I have to first write better documentation for extending SysFeatures. At the moment, it describes the changes not well.

Please add it to the changelog, I forgot to mention that.

There is also the architecture docs in `doc/archs` and the architecture page in the wiki. You can do that or leave it to me. It's just copying the pages and changing some identifiers.

vxirau:
Not sure how you want the wiki done! I believe the rest is OK :)

ipatix:
I am currently in the process of refactoring code to enable compiler warnings. During this I noticed a warning in Emerald Rapids code:

https://github.com/RRZE-HPC/likwid/blob/81096484e8b79be4520a7377608a2af7f66b7736/src/includes/perfmon_emeraldrapids.h#L262

@vxirau Is it possible that `counter_map[index].index <= CBOX55` is supposed to be `counter_map[index].type <= CBOX55`?

vxirau:
Hi @ipatix it most definitely can be! I mostly copied it directly from the sapphire rapids code, so this issue will be present there as well.

Sorry for the oversight!

ipatix:
> Sorry for the oversight!

No problem! You are right, it is indeed also present in the Sapphire Rapids code. Thanks for the hint.

PR #670, opened 2025-03-24
UPI flits events added for Sapphire Rapids for TXL_ALL_NULL, TXL_NON_…

guillemguerrero:
Hello! I was using likwid on an SPR system and needed to use some events that were not there. I've added them just in case someone else needs them aslo in the future. I wasn't sure if I should increase the version (5.4.1) or if you're changing it only when you have changes accross many archs.

### DIFF
--- a/src/includes/perfmon_sapphirerapids_events.txt
+++ b/src/includes/perfmon_sapphirerapids_events.txt
@@ -4,13 +4,13 @@
 #
 #      Description:  Event list for Intel Sapphire Rapids
 #
-#      Version:   5.4.1
-#      Released:  09.12.2024
+#      Version:   <VERSION>
+#      Released:  <DATE>
 #
 #      Author:   Thomas Gruber (tr), thomas.roehl@googlemail.com
 #      Project:  likwid
 #
-#      Copyright (C) 2024 RRZE, University Erlangen-Nuremberg
+#      Copyright (C) 2016 RRZE, University Erlangen-Nuremberg
 #
 #      This program is free software: you can redistribute it and/or modify it under
 #      the terms of the GNU General Public License as published by the Free Software
--- a/src/includes/perfmon_sapphirerapids_events.txt
+++ b/src/includes/perfmon_sapphirerapids_events.txt
@@ -4,13 +4,13 @@
 #
 #      Description:  Event list for Intel Sapphire Rapids
 #
-#      Version:   <VERSION>
-#      Released:  <DATE>
+#      Version:   5.4.1
+#      Released:  09.12.2024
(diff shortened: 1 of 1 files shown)


TomTheBear:
Hello. Thanks for the PR. It is definitely valuable. Did the unit work reliably for you? All (up to) four UPI channels and all counters? There is a hardware bug about these units and I added [a workaround](https://github.com/RRZE-HPC/likwid/pull/639/commits/f4a0ba8ddf3fdad949389ab3bf2c1c32a2a8668c) but it only worked for `UPI*C0` and `UPI*C2` (see [`UPI`](https://github.com/RRZE-HPC/likwid/blob/master/groups/SPR/UPI.txt) performance group). I have an idea what could be the reason but didn't have time to test it yet.

The version, date and also the copyright line is updated ahead of releases with the `perf/set_license.pl` script. So please, revert it back to `<VERSION>`, `<DATE>` and the copyright year (although that does not matter, the scripts replaces any years).

guillemguerrero:
No problem! All three lines have been reverted to their original state. 

Regarding the hardware bug you mention, I'm a bit unsure about the effects it had... Could you provide a bit more information on the behavior it had? We could read the counters, but we assumed that they contained valid information. By checking the workaround we think we can infer what we think the workaround is, but it would be nice to have a deeper explanation to know how to determine if the measurements were correct or not.

TomTheBear:
If they work for you, you already use the workaround.

I'm currently in Barcelona without all my stuff but I try to summarize it. There is a public intel document mentioning this eratum for SPR but I don't find it spontaneously. 

In order to find the Uncore units for SPR and newer Intel systems, you have to run the so-called Uncore discovery mechanism. For each unit, it returns info like base address of the unit in memory, offsets to the first config and counter registers, register width and the number of counter registers. For the UPI and M3UPI units, the base addresses are wrong and you have to look them up as a workaround. The Linux kernel has the Dame workaround, there I saw it the first time.

The issue with the not-working odd numbered counters is somewhat related. You get the offset to the first config register and the register width but the register width is only valid for the counter registers. So, it is unclear how wide the config registers are. I know already for some other units (like MBOX and HBM) that the config registers are only 32 bit. My guess is that the UPI config registers are also only 32 bit.

I hope this is somewhat clear.

guillemguerrero:
It sure does, thanks for the explanation!

TomTheBear:
There was a bug in the uncore discovery mechanism in LIKWID. You should upgrade to https://github.com/RRZE-HPC/likwid/commit/c2884e94289dd030b6d87afa313409192ae3608d

PR #675, opened 2025-04-17
Adding support for llvm clang on arm

fbarbari:
This is an attempt to make likwid compile on ARM with llvm clang. Right now I am testing on a Neoverse-N1 with llvm@19.1.7.
I added the config file for CLANGARMv8 merging the ones for ARMCLANG, CLANG and GCCARMv8.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -132,6 +132,19 @@ OBJ := $(filter-out $(BUILD_DIR)/access_x86_translate.o,$(OBJ))
 else
 OBJ := $(filter-out $(BUILD_DIR)/loadDataARM.o,$(OBJ))
 endif
+ifeq ($(COMPILER), CLANGARMv8)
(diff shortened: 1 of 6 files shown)


fbarbari:
To be clear, right now the PR is not ready but I chose to open it early since I am facing some compilation errors I do not quite understand how to fix, so I'm asking for help.

For example:

```
/home/fbarbari/likwid/src/topology_cpuid.c:75:9: error: call to undeclared function 'CPUID'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
   75 |         CPUID(eax, ebx, ecx, edx);
      |
```
Shouldn't the CPUID macro be defined as "nothing" in `src/includes/cpuid.h` on arm?

Thanks in advance

PR #688, opened 2025-06-26
Support for AMD Zen5 (Family 0x1A, Model 0x2)

TomTheBear:
WIP

- [x] Topology detection
- [x] Energy module
- [x] Perfmon counter list
- [x] Perfmon functions
- [ ] Perfmon events
- [ ] Perfmon with perf_event
- [ ] Perfmon performance groups
- [ ] Sysfeatures

Some initial events are added. Sysfeatures should work as well but RAPL needs to be checked.


TomTheBear:
This is a message often showing up in syslog. It is a annoying message by the kernel devs because they want you to use `perf_event` instead of other solutions. But this is just a message, access should still work.

> Update for Linux kernel 5.9 and newer: With Linux 5.9, the msr kernel module got some security fixes. The major change for LIKWID is, that now all MSR are non-writable by default. In order to change that, you have to change the boot options of your operating system to contain `msr.allow_writes=on` to enable writes again. This affects only ACCESSMODE=direct and ACCESSMODE=accessdaemon. If you use the perf_event backend, you don't have to change anything.

With `msr.allow_writes=on` the messages should disappear.

TomTheBear:
Support for Zen5 is complete!

nolta:
Trying this out, and it appears to work. Thanks!

ohm314:
Not sure this is the right place to report this, but since I guess you're testing your Zen5 PR, it sorta makes sense:

I'm running on a Zen5 platform and thought I give your PR a try. In my case likwid-perfctr crashes reporting an FPE:

```
likwid-perfctr -V 3 -C 0 -g FLOPS_SP,MEM likwid-bench -t peakflops_sp_avx512_fma -w S0:512kB
DEBUG - [hwloc_init_cpuInfo:360] HWLOC CpuInfo Family 26 Model 2 Stepping 1 Vendor 0x0 Part 0x0 isIntel 0 numHWThreads 8 activeHWThreads 8
... more debug output ...
--------------------------------------------------------------------------------
CPU name:       AMD EPYC 9J45 128-Core Processor
CPU type:       AMD K1A (Zen5) architecture
CPU clock:      2.70 GHz
CPU family:     26
CPU model:      2
CPU vendor:     0
CPU part:       0
CPU short:      zen5
CPU stepping:   1
CPU features:   FP MMX SSE SSE2 HTT MMX RDTSCP SSSE FMA SSE4.1 SSE4.2 AES AVX RDRAND AVX2 AVX512 RDSEED SSE3
CPU arch:       x86_64
PERFMON supports Uncore:        1
--------------------------------------------------------------------------------
[likwid-pin] Main PID -> hwthread 0 - OK
DEBUG - [HPMinit:114] Adjusting functions for x86 architecture in daemon mode
DEBUG - [perfmon_init_maps:933] Initialize maps for family 0x1A model 0x2 stepping 0x1
DEBUG - [perfmon_init_maps:1661] Creating runtime counter map
Floating point exception (core dumped)
```



Happy to create an issue if that's more appropriate.

ohm314:
ah.. querying `0x80000021` and checking bit 17, I guess CpuidUserDis is not set. I'm running in a cloud VM instance, i guess that's normal? Sorry.. don't remember I ever dug that deep in this particular corner :)

TomTheBear:
Cloud instances are often operated with a limited feature set. Sometimes you are lucky with a bare-metal instances. You can potentially monitor memory traffic of other jobs running on the same CPU socket. The same code works fine on my Zen5 nodes, so I guess your cloud vendor turned it off.

Can you please update your installation with the latest fixes and try again. Thx.

ohm314:
Yeah, I figured that I can't access to much with the VM instance I have at hand.
On the upside, I've checked out your latest commits and now all runs smoothly, so at least one corner case was caught :)
I don't have access to a baremetal at this moment, but if I manage to get hold of one I can quickly rerun there and update here if there is something interesting to share.
Thanks for your work!

TomTheBear:
Thanks for testing and the positive feedback.

ohm314:
I got access to a baremetal instance and I'm happy to report that the PR works like a charm

ohm314:
Found a tiny typo to fix:

```patch
diff --git a/groups/zen5/FLOPS_SP.txt b/groups/zen5/FLOPS_SP.txt
index f1b279c9..c68c7ca8 100644
--- a/groups/zen5/FLOPS_SP.txt
+++ b/groups/zen5/FLOPS_SP.txt
@@ -15,7 +15,7 @@ Runtime (RDTSC) [s] time
 Runtime unhalted [s]   FIXC1*inverseClock
 Clock [MHz]  1.E-06*(FIXC1/FIXC2)/inverseClock
 CPI   PMC1/PMC0
-SP [MFLOP/s]   1.0E-06*(PMC2+PMC2)/time
+SP [MFLOP/s]   1.0E-06*(PMC2+PMC4)/time

 LONG
 Formulas:
```

Issue #466, opened 2022-05-22
[FeatureRequest] Add TMA_BACKEND, TMA_BE_MEMORY and TMA_BE_CORE counter group to likwid-perfctr


ibogosavljevic:
**Is your feature request related to a problem? Please describe.**
I am frustrated when doing top-down analysis with Intel's VTUNE. The tool us cumbersome and affects the results. Also I cannot limit to the code I am interested in, only whole functions LIKWID already has TMA counter group, but it should move further with additional groups which are deep

**Describe the solution you'd like**
I would like to be able to do the same analysis with LIKWID. For the beginning, I would like two additional groups: TMA_BE_MEMORY and TMA_BE_CORE. Here is the possible output:

TMA_BE_MEMORY
```
L1 Bound: 
L2 Bound:
L3 Bound:
DRAM Bound:
Store Bound:
``` 

TMA_BE_CORE
```
Divider: 19.7% of Clockticks
Port Utilization: 43.3% of Clockticks
    Cycles of 0 Ports Utilized: 19.5% of Clockticks
    Cycles of 1 Port Utilized: 7.1% of Clockticks
    Cycles of 2 Ports Utilized: 6.0% of Clockticks
    Cycles of 3+ Ports Utilized: 14.2% of Clockticks
```

**Additional context**
You will probably need this to implement it:

This performance group measures cycles to determine percentage of time spent in
front end, back end, retiring and speculation. These metrics are published and
verified by Intel. Further information:
Webpage describing Top-Down Method and its usage in Intel vTune:
https://software.intel.com/en-us/vtune-amplifier-help-tuning-applications-using-a-top-down-microarchitecture-analysis-method
Paper by Yasin Ahmad:
https://sites.google.com/site/analysismethods/yasin-pubs/TopDown-Yasin-ISPASS14.pdf?attredirects=0
Slides by Yasin Ahmad:
http://www.cs.technion.ac.il/~erangi/TMA_using_Linux_perf__Ahmad_Yasin.pdf
The Intel Icelake microarchitecture provides a distinct register for the Top-Down Method metrics.

TomTheBear:
[...]

PR #690, opened 2025-07-09
 Initial Support for Reading Intel TMA Events via PMCs on Sapphire Rapids

vxirau:
I’ve been interested in supporting Intel’s Top-Down Microarchitecture Analysis (TMA) events on Sapphire Rapids, especially through direct PMU programming. 

While there are known challenges in accessing TMA registers, I saw some were described in [[RRZE-HPC/likwid#466](https://github.com/RRZE-HPC/likwid/issues/466)](https://github.com/RRZE-HPC/likwid/issues/466). Intel documentation suggests this should be possible.

According to Intel’s Software Developer Manual:

 [[Intel SDM Vol. 3B – Section 20.3.10.1.1, Page 291](https://cdrdv2-public.intel.com/671427/253669-sdm-vol-3b.pdf)](https://cdrdv2-public.intel.com/671427/253669-sdm-vol-3b.pdf)

intel

There are dedicated TMA registers not only for L1 metrics but also for L2 metrics. This would translate from TMA0...7. I've added these events to the events file, even though they cannot be used as of now.

---

### What This PR Adds

Since perf can indeed measure the TMA metrics, I tried checking how, and I've added

* A new group configuration to directly read TMA-related events using PMCs.
* Manual setup of the following event/umask combinations, which are used internally by `perf stat`:

```
event=0x00,umask=0x80  → topdown-retiring
event=0x00,umask=0x81  → topdown-bad-spec
event=0x00,umask=0x82  → topdown-fe-bound
event=0x00,umask=0x83  → topdown-be-bound
event=0x00,umask=0x84  → topdown-heavy-ops
event=0x00,umask=0x85  → topdown-br-mispredict
event=0x00,umask=0x86  → topdown-fetch-lat
event=0x00,umask=0x87  → topdown-mem-bound
```

These are now readable via standard PMCs, and initial testing shows the reported values are consistent with what `perf` produces.

---

### Status and Limitations

Although the current implementation **works via PMCs**, it does **not yet use the dedicated TMA registers** (`MSR_PERF_METRICS`). Getting access to those registers would be ideal, as it would allow:

* The 8 TMA events to be measured in parallel via the TMA engine
* The 8 general-purpose PMCs to remain available for other events
* Without impacting existing uncore measurements.

This is currently possible in `perf`, which can leverage the TMA registers separately from PMCs, effectively doubling the number of metrics collected in a single run without requiring multiplexing.

At present, LIKWID relies on PMCs and does not use the TMA registers. If support for enabling and reading those registers were added (e.g., via `MSR_PERF_METRICS_ENABLE`), it could improve measurement density and efficiency, especially for workloads where TMA + other core events + uncore events are needed concurrently.

---

### Implementation Notes

Intel’s documentation indicates that enabling `MSR_PERF_METRICS` requires:

* A per-core reset (`MSR_PERF_METRICS_ENABLE = 0`) issued once
* A per-logical-thread enable (`MSR_PERF_METRICS_ENABLE = 1`) on each thread

This sequence is needed to activate the TMA engine and ensure non-zero results in the associated MSRs. Without this, values reported from `MSR_PERF_METRICS` (and possibly other TMA-related registers) will remain zero.

This PR attempts to implement that logic too, but I was not successfull in reading the TMA registers. 

Currently, LIKWID supports reading uncore and core counters together. If TMA registers can be accessed in parallel with PMC events (without multiplexing), it may be possible to integrate them similarly using the existing infrastructure.

---

### Summary

This PR:

* Adds support for measuring TMA events using standard PMCs
* Demonstrates consistency with `perf` output
* Attempts to expand TMA register support
* Opens the door to broader TMA + PMC + uncore measurement combinations in a single run

My experience with the LIKWID codebase is limited, but I am very familiar with hardware performance counters and low-level register behavior. I have root access to an Emerald Rapids system and user access to a Sapphire Rapids system, and I’m happy to run tests or assist with validation if that helps move this forward.

Thanks!


TomTheBear:
Thanks for the PR. But, unfortunately, I do not totally understand why you did it this way.

The SPR code already contained the topdown level 1 events already through the `TMA*` counters. These counters are reading the `PERF_METRICS` MSR. You added the additional events for SPR as `TMA4`to `TMA7`. For the accessdaemon and direct access mode, this is enough :
```
|        RETIRING       |   TMA0  |     0.0941 |
|    BAD_SPECULATION    |   TMA1  |     0.1176 |
|     FRONTEND_BOUND    |   TMA2  |     0.4588 |
|     BACKEND_BOUND     |   TMA3  |     0.3373 |
|       HEAVY_OPS       |   TMA4  |     0.0314 |
|     BR_MISPREDICT     |   TMA5  |     0.1137 |
|     FETCH_LATENCY     |   TMA6  |     0.3765 |
|       MEM_BOUND       |   TMA7  |     0.1804 |
```

The enable bit is in `MSR_PERF_GLOBAL_CTRL`, I could not find `MSR_PERF_METRICS_ENABLE=0x3F0` in the documentation for SPR. It might be available in later generations or the P-cores of these P/E/Elight systems, I haven't checked them.

But as you have seen in the `counters.h` file, I have not been successful using them with `perf_event`. The reason for it is likely that the events are added to LIKWID without the right configuration settings. In the current event config `umask=0x0` for all TMA events. You listed the right configurations. Unfortunately, I cannot test `perf_event`. While the TMA events are listed in `/sys/devices/cpu/events`, `perf` only returns `not supported` on my system.

So, in my opinion, add the umask settings to the `TMA*` events is a better way. Then all `PMC*` can be used for other measurements.

vxirau:
Hello @TomTheBear,

Thanks for the clarification, you're absolutely right that the TMA0–TMA7 counters already work via the PERF_METRICS MSR through accessdaemon. I tried the PMC fallback because I believed it was the supported approach, but it wasn’t working on our test machines, and I was unaware that the accessDaemon path allowed for correct reading.

Since the direct MSR approach via accessdaemon is the right way forward, I’ll focus on getting accessdaemon support up and running on our end instead of the PMC workaround.

Regarding TMA4–7, I can update this PR to:
* Add the correct settings for TMA4..TMA7 to the events file
* Propagate those changes to architectures beyond Sapphire Rapids (Emerald, Granite, ...)

Let me know if you’d like me to push that update here, or if you’d prefer to handle it yourself.

Thanks!

TomTheBear:
Can you please check whether https://github.com/RRZE-HPC/likwid/pull/694 works for you? Thanks

TomTheBear:
The PR #694 is now in the master branch

PR #694, opened 2025-07-11
Add Topdown level 2 metrics for SPR, EMR and GNR

TomTheBear:


TomTheBear:
Issues with topdown events and `perf_event`: While the individual topdown events are documented officially by `perf_event`, they do not work individually. 

```
topdown-bad-spec  topdown-be-bound  topdown-br-mispredict  topdown-fe-bound
topdown-fetch-lat  topdown-heavy-ops  topdown-mem-bound  topdown-retiring
```
with e.g. `topdown-retiring` = `event=0x00,umask=0x80`

All topdown events have to be in one socalled event group even if you do not want to measure them. LIKWID's `perf_event` backend currently does not use event groups.

PR #710, opened 2025-12-13
Standardize help text

iamazeem:
- Fixed minor typos
- Fixed formatting issues
- Used consistent format for examples i.e.
  - `# Description`
  - `$ command`
- Replaced `print` with `print_stdout` for consistency across all applications
- Marked `version` and `usage` functions as `local`
- Built, tested, and verified locally on Ubuntu 22.04

Signed-off-by: Azeem Sajid <azeem.sajid@gmail.com>


ipatix:
Hi, thanks for the PR.

While we generally are not against the use of AI generated code, please review your code changes before submitting your PR. If you are not sure at the time of creation create a draft.

In your case specifically the PR makes the output look more consistent in some places, but makes it worse in others. Please fix this.

Also, I am not particularly a fan of describing examples like this:
```
# Examplanation
$ echo some shell command
```
instead of the previous

```
Examplanation
$ echo some shell command
```
The `#` makes the description indented to the same level as the command, which (to me) decreases visual contrast between the two. Also the `$` is not shell syntax, so it's not like we need a 'syntactical' `#`.

ipatix:
Thanks, but can you please go over *all* tools in src/applications and make sure that the style is consistent? If the point of this PR is to standardize the help text, and you change the usage of `/` and `,` (which is totally fine), please do so consistently.

PR #711, opened 2025-12-19
New Groups and Events for GNR + SPECI2M Fix for GNR, EMR and SPR

paudiaz99:
Hi!

I've been working on measuring SpecI2M traffic on GNR, EMR and SPR, and I noticed that some events were missing and the SPECI2M group was not capturing the traffic correctly.

In order to fix this, I added the event TOR_INSERTS_IA_SPECI2M (Present in https://perfmon-events.intel.com/platforms/graniterapids/uncore-events/uncore/) to GNR, EMR and SPR event list.
For the GNR platform I identified some missing CHA and IMC counters, so I also added them.

Additionally, I modified the SPECI2M group of these three platforms to correctly measure SPECI2M traffic by using the previously mentioned event, and added TLB_DATA and TLB_INSTR groups to GNR. Finally, I modified src/perfgroup.c to allocate more memory to accommodate the long lines in new "groups/GNR/SPECI2M.txt" file and prevent an overflow.

I used this version to measure the SpecI2M traffic on the mentioned platforms, so I thought it would be nice to integrate it on the main branch:). In case of the SpecI2M group for the ICX platform, I don't have access to this platform to test if the current event is correct or not. 

Thanks!


TomTheBear:
I cherry-picked the commit also to the v5.5 branch. I'm currently preparing 5.5.1, so it will be part of this release already.

Issue #634, opened 2024-09-11
[BUG] Extremely long wall clock time when using the marker API in OpenMP code


tdollenb:
When using the marker API along with OpenMP I observed a huge runtimes with `likwid-perfctr`. I was able to reproduce the issue with the [tutorial on markers in C](https://github.com/RRZE-HPC/likwid/wiki/TutorialMarkerC). The tutorial code runs 40 times longer on my system when using `likwid-perfctr -m` compared to running without `-m`. However, the measurements seem correct, the application seems to be halted a lot.

* System: Intel SPR (Xeon(R) Platinum 8468), 2x48Cores
* OS: LINUX, ROCKY 8.10
* LIKWID version is 5.3.0 with additional bugfixes up to commit ID -d8fea29 deployed with Easybuild.
* Run command: `likwid-perfctr -g BRANCH -C 0-95 -m ./a.out`. The code is the same as in the marker tutorial.
* Compiled with Intel(R) oneAPI DPC++/C++ Compiler 2023.1.0 (2023.1.0.20230320)
* Compilation command: `icx -fiopenmp -O3 -DLIKWID_PERFMON -llikwid marker_tutorial.c`

Output with `-m`:
[likwid_dump.txt](https://github.com/user-attachments/files/16962778/likwid_dump.txt)

TomTheBear:
The code on the tutorial page is just a demo how the code structure should look like when adding the MarkerAPI. It is probably running too short for anything helpful. According to your output, the OpenMP threads had an average runtime of 0.0007s.

The most runtime is spent setting up the counting infrastructure at `LIKWID_MARKER_REGISTER` (or the first `LIKWID_MARKER_START` if no `LIKWID_MARKER_REGISTER` was done before) for each OpenMP thread. This happens only once and applications with significant runtime should not be affected much. Your selected `ACCESSMODE=accessdaemon` has the largest startup time.

tdollenb:
I used the tutorial application to reproduce it with something small, but I can see the issue with larger, long running applications aswell. The sparse matrix vector multiplication code I observed it with is below. I tried to use `LIKWID_MARKER_REGISTER` aswell but it had no effect on the runtime:

```
LIKWID_MARKER_INIT;
#pragma omp parallel num_threads(tOptions->iNumThreads)
{
	LIKWID_MARKER_THREADINIT;
	LIKWID_MARKER_REGISTER("Compute");
}

// take the time: start
t1 = omp_get_wtime();

[...]

PR #728, opened 2026-03-16
Add the ability to release and reacquire performance counters in the perfmon API

dassschaf:
Hi everyone,

this PR adds the ability to release and (re-)acquire (`perfmon_acquireCounters`, `perfmon_releaseCounters`) the performance counters in the API, extracted from the existing functions `perfmon_init` and `perfmon_finalize`.
This addition has several benefits in the context of performance monitoring:
- A tool to monitor the performance in the background can acquire and release the counters for use so that another application does not have to wait for the finalization upon creation of a lockfile or any other mechanism to pause the monitoring tool for a job requiring access for the counters so the force flag does not need to be used.
- Acquire/Release is much faster than fully initializing/finalizing (around 5-10ms vs. around 3s on the cluster here in Aachen; see attached code example below). The repeated initialization for monitoring the performance of the cluster in the background causes measurable overhead (such as in issue #634) and can be avoided with merely acquiring and releasing the counters.

I refrained from running `make format` on the changes as it changed virtually every line in the two changed files.

To try the [attached code example](https://github.com/user-attachments/files/26023358/likwid_example.zip):
```
$ g++ -std=c++20 -g -O3 -llikwid -o likwid_example.exe likwid_example.cpp
$ ./likwid_minimal.exe
Starting measurements...
....................
20 times initializing/finalizing access took 54471 ms, avg.: 2723 ms 

....................
20 times acquiring/releasing access took 112 ms, avg.: 5 ms 
```

### DIFF
--- a/src/includes/likwid.h
+++ b/src/includes/likwid.h
@@ -957,8 +957,9 @@ Acquires all performance monitoring counters previously released by
 perfmon_releaseCounters(), takes the same parameters as perfmon_init(...).
 @param [in] nrThreads (Number of Threads)
 @param [in] threadsToCpu (List of CPUs)
+@return error code (0 on success, -ERRORCODE on failure)
 */
-extern void perfmon_acquireCounters(int nrThreads, const int* threadsToCpu) 
+extern int perfmon_acquireCounters(int nrThreads, const int* threadsToCpu) 
     __attribute__((visibility("default")));
 /*! \brief Start performance monitoring counters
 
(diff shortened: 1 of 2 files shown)


chriswasser:
Thanks for your evaluation 👍 It sounds like, you see no large road block for integrating this feature and extending the public API with the two proposed `perfmon_acquireCounters` and `perfmon_releaseCounters` functions. Let us know how you would like to continue with this PR and whether you need any additional information or tests from our side 🤓 Greetings

chriswasser:
Hi @TomTheBear,

Is there any status update on getting this feature integrated into main? Just asking again, since we would love to see this feature included in the next LIKWID release as it eliminates the potential friction of our monitoring application with user applications.

Thanks in advance and greetings 👋

TomTheBear:
Unfortunately, I'm still not sure.

PR #753, opened 2026-06-11
Fix build option and access_daemon_main() for likwid-bridge

sven-hansen:
This PR fixes two regressions in recent commits that broke the build for the container helper likwid-bridge. I stumbled upon this when trying to build 5.5.2rc1. 

The problem with building the bridge in the first place was introduced with commit 9fdaa8b1fd356436fccf88cb385807d0291c3429 because now the variable `CONTAINER_HELPER` is defined twice, once as a build flag and once as the path definition for the resulting daemon. This PR renames the build flag to `BUILD_CONTAINER_HELPER` to match the naming of similar options (e.g., for the app daemon). 

Additionally, #738  rewrote the code portions for failed exec calls but left references to the undefined `ret` in `access_daemon_main()`:
https://github.com/RRZE-HPC/likwid/blob/1142c8aad076e1b5c726697dc4eb29b70dc37a8c/src/bridge/bridge.c#L50-L55

Since the return value has no merit here, I removed it altogether so the different code paths look identical now. 🙂


TomTheBear:
Thanks

